PR #21, opened 2021-07-20
Add and read shopping list items 

### LINKED ISSUE #4: 4. As a user, I want to add a new item to my shopping list so I can start recording purchases.
A shopping list item consists of the following data points:

* Name of item
* How soon are you likely to buy it again?
  * Soon (in the next 7 days)
  * Kind of soon (in the next 14 days)
  * Not soon (in the next 30 days)
* Last purchased date

AC:

* User is presented with a form that allows them to enter the name of the item and select how soon they anticipate needing to buy it again (Soon, Kind of Soon, Not Soon)
* When the user submits the form, the item is saved to the database, associated with the user’s token
* Along with the item name, an integer corresponding to the estimated number of days until next purchase is saved: `7` for “Soon”, `14` for “Kind of Soon”, and `30` for “Not Soon”
* The last purchased date should be set to `null` initially (i.e. you can create an item without purchasing it)
* Item names should be displayed as a simple, unordered list on the list view after they’re added

Notes:

* The add item form has the potential to be a lot of scope. The wireframe https://github.com/the-collab-lab/smart-shopping-list/blob/main/_resources/wireframes/add-item.png has several notes related to accessibility and UX.
* Eventually, the list will look like the following wireframe: https://github.com/the-collab-lab/smart-shopping-list/blob/main/_resources/wireframes/list-populated.png
* It's good for your team to look at the wireframe now, but be sure not to try to implement all of the visible functionality at once. There will be future stories to add all of those features.
* For this story, it's sufficient just to get a simple list of the items displaying.

anderswift:
## Description

This PR adds a component AddItemForm to add shopping list items, and components ShoppingList and ShoppingListItem to display shopping list items. 

We also started to arrange things inside the ListView and AddItemView components to match the wireframe (including adding a barebones Header component), and added CSS classes for later to the files we were working on. 

### AddItemForm
We created a form with controlled components, which means the form fields update a state variable (formValues) every time the user changes anything. On form submission, the current values are formatted for the database, and defaults for the last purchase date and number of purchases are also set. The form relies on a listId value passed down from the App component, which is used to add the item to the correct list (the one corresponding to the user token. 

### ShoppingList and ShoppingListItem
Again relying on the listId value passed down from the App component, we use a useCollection Firebase hook to fetch the items from the correct list document in Firestore. Since it's going to be more complex later on, we separated out the list item into its own component to keep things neat and tidy.

### Changes to App component
In order to have a listId to pass to our new components and use when reading and writing to Firestore, we added a useEffect hook to App.js to, on initial mount, run a check for the localStorage token and fetch the corresponding List ID from Firestore.

## Related Issue

Closes #4

## Acceptance Criteria

- [x] User is presented with a form that allows them to enter the name of the item and select how soon they anticipate needing to buy it again (Soon, Kind of Soon, Not Soon)
- [x] When the user submits the form, the item is saved to the database, associated with the user’s token
- [x] Along with the item name, an integer corresponding to the estimated number of days until next purchase is saved: 7 for “Soon”, 14 for “Kind of Soon”, and 30 for “Not Soon”
- [x] The last purchased date should be set to null initially (i.e. you can create an item without purchasing it)
- [x] Item names should be displayed as a simple, unordered list on the list view after they’re added

## Type of Changes

### Before
image
image

### After
image
image

1. Pull down this branch and check it out.
2. Run npm install to update dependencies.
3. Run npm start to start local server.
4. Add our test token (token: three word token) manually to localStorage to allow the app to connect with our test list in Firestore.
4. Fill out the "Add Item" form with a new item and submit. Check the list view and [the database](https://console.firebase.google.com/project/tcl-26-shopping-list/firestore/data/) to make sure it added.


SheilaRMoore (commented):
I'm wondering if for the local storage if we can maybe put the functions for that in one centralized place, since we are overlapping. Maybe we could add a localstorage file to lib, and give it 2 export default functions. One function could be for checking for the token, and returning to some default value if there isn't a token, and the other could be for adding a value/key/token. This way we could just import it to App.js. and Home.js.  Maybe we can discuss as a group. I like the idea of the listId and how it was made reusable. Overall I think you guys did a great job.

SheilaRMoore (approved):

src/components/ShoppingList/ShoppingList.js
```diff
@@ -0,0 +1,35 @@
...
+      )}
+
+      {error && (
+        <div className="shopping-list__notice notice notice_type_error">
+          Error
+        </div>
+      )}
+
+      {!loading && listItems && (
+        <ul className="shopping-list__list">
```
anderswift, line 25: I know. But trying to stick using BEM conventions and component names for 'block' css classes, this seemed the best BEMy name for a class to style the ul inside the ShoppingList component.

src/components/AddItemForm/AddItemForm.js
```diff
@@ -0,0 +1,115 @@
...
+      <input
+        className="add-item-form__text-field text-field"
+        type="text"
+        id="itemName"
+        name="itemName"
+        aria-describedby="itemNameHint"
+        value={formValues.itemName}
+        onChange={handleChange}
+        maxLength="100"
+        required
```
anderswift, line 52: We added in these basic limits on the field, but I'd ideally like to do some form validation with feedback to the user. That was the intention of the empty span below, but we didn't have a chance to finish.

src/components/AddItemForm/AddItemForm.js
```diff
@@ -0,0 +1,115 @@
...
+    newItem.lastPurchaseDate = null;
+    newItem.numberOfPurchases = 0;
+
+    try {
+      await db.collection(`lists/${listId}/items`).add(formValues); // add item to Firestore database
+      setFormValues(defaultFormValues); // after saving to database, reset form values to defaults
+    } catch (err) {
+      console.log(err);
+    }
+  };
```
anderswift, line 33: Would be good to handle giving feedback to the user upon success/failure in here as well.

anderswift:
> I'm wondering if for the local storage if we can maybe put the functions for that in one centralized place, since we are overlapping. Maybe we could add a localstorage file to lib, and give it 2 export default functions. One function could be for checking for the token, and returning to some default value if there isn't a token, and the other could be for adding a value/key/token. This way we could just import it to App.js. and Home.js.

My instinct would be if I'm doing something with just a couple of lines and I won't need to repeat those instructions somewhere else, I probably wouldn't bother to make it a separate function (thinking it would be clearer to follow what was happening if it was kept in one file). But if we are linking our code together, and therefore adding in fetching/creating a list in the database along with getting/setting the token, I think I agree that it would make things cleaner to name functions and stick them in /lib.

What do you all think of something like this, in /lib as Sheila suggested?
1. A function to generate a token, save it in localStorage, create a list with the token in Firestore, and return the document ID of that list (well, a Promise that would resolve to the list ID or an error)
2. A function to check for a token in localStorage, check for a list associated in the database, and return again a Promise that would resolve to the list ID, maybe null if no token exists, or an error (invalid token, connection problem)

src/components/AddItemForm/AddItemForm.js
```diff
@@ -0,0 +1,115 @@
...
+
+  // on form submit, add user input and other default values to database
+  const handleSubmit = async (event) => {
+    event.preventDefault();
+
+    const newItem = formValues;
+    newItem.purchaseInterval = parseInt(newItem.purchaseInterval); // convert purchaseInterval to integer
+    newItem.lastPurchaseDate = null;
+    newItem.numberOfPurchases = 0;
```
nick-zanetti, line 25: I really like the way you set this up!

src/components/AddItemForm/AddItemForm.js
```diff
@@ -0,0 +1,115 @@
...
+
+        <input
+          type="radio"
+          className="add-item-form__radio"
+          id="kindaSoonOption"
+          name="purchaseInterval"
+          value="14"
+          onChange={handleChange}
+          checked={formValues.purchaseInterval === '14'}
+        />
```
nick-zanetti, line 83: For these `purchaseInterval` items, is it possible to simply set them to an actual number? Then we would not have to do this up above: 

`newItem.purchaseInterval = parseInt(newItem.purchaseInterval); // convert purchaseInterval to integer`

nick-zanetti (approved):
Very nicely done overall! Thank you for including so many code comments as this PR had a very big scope and a lot going on. It reminded me that I need to do more commenting myself on my own code. I think this fulfills all the AC as it currently is, but I did have some additional comments throughout.

src/App.js
```diff
@@ -1,33 +1,61 @@
...
+          // for some reason, extracting this array seems necessary to avoid errors?
+          const documents = querySnapshot.docs;
+
+          // check if the expected data is returned
+          if (
+            Array.isArray(documents) &&
+            documents.length &&
+            typeof documents[0] === 'object'
+          ) {
+            setListId(documents[0].id); // save the list id for later
```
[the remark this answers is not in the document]
Cjstoney: I really like how thorough this check is! @nick-zanetti  I believe the check for length is to ensure that we are not getting back an empty array. It could be argued that the next line makes that check as well by assuming that there is a value at the index of 0, but personally I like how explicit line 35 is about the check

src/components/AddItemForm/AddItemForm.js
Cjstoney: This is a great instinct to have. I would gently suggest that when you come across a situation where you run out of time, create an issue to address the item that you would like to continue on. This way no one comes across this span while working on another feature which could lead to merge conflicts.

Cjstoney (approved):
This is looking great. I am really enjoying how forward thinking your code is. Great job!

src/App.js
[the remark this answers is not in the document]
anderswift: 1. Is it an array?
2. Is there anything in the array? (length != 0)
3. Is the first (and probably only) thing in the array an object

...to be even more safe could probably add:
4. Does the object even have an id property?

The thinking is doing these checks, in this order, will hopefully prevent getting errors if something doesn't exist in the database or there's some other problem with results returned. Just straight up referring to `documents[0].id` if I didn't get back data in the format I'm expecting maybe not go over well. Though honestly, I haven't worked enough with firebase or had time yet to test different possibilities and learn the format of the result I'd get in those cases. 

This is likely a habit I'm in from coding in PHP, where if you make an assumption like that and it doesn't pan out, your entire page might be kaput (or at best of visible errors), so checks like this limit the failure to one area where you write in a fallback. Interested to hear from mentors if this is overkill in JavaScript?

src/components/AddItemForm/AddItemForm.js
anderswift: Yes, @Cjstoney, Connie and I talked about that and opening an issue for form validation is on my to do list!

src/App.js
```diff
@@ -1,33 +1,61 @@
...
+    const token = localStorage.getItem('token');
+
+    if (token) {
+      // find the list in Firestore associated with the stored token from local storage
+      db.collection('lists')
+        .where('token', '==', token)
+        .get()
+        .then((querySnapshot) => {
+          // for some reason, extracting this array seems necessary to avoid errors?
+          const documents = querySnapshot.docs;
```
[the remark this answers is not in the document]
anderswift: @nick-zanetti The story here is that we tried to use something like `querySnapshot.docs[0]`, and `querySnapshot.docs[0].id` just to check out what we were getting back from Firestore, and got weird results. Then when I stuck it in a new variable `documents` it worked fine. Now that I go back and try again, `querySnapshot.docs[0].id` does work though? So I'm not sure what was going on there but you can ignore that comment and we'll take it out next revision!

In general, we were a bit pressed for time this week so pretty much ran with the first thing that worked when it came to Firebase. Would love some feedback from @yenly @abowler2 @Cjstoney on the Firestore stuff! For example, I feel like there must be a better or more concise way to get just one document matching an ID. (I definitely know how to do that with multiple other databases, but Firestore is new to me and the documentation seems a tad short on examples.)

src/components/AddItemForm/AddItemForm.js
anderswift: Thanks @nick-zanetti! I know not super relevant for our project, but I like to do form values in a single state object with a single onChange if I can, because it makes it really painless if you ever need to change the form around and add/remove fields. Plus nice and concise!

src/App.js
```diff
@@ -1,33 +1,61 @@
...
+    }
+  }, []);
+
   return (
     <Router>
       <div className="App container">
         <Switch>
           <Route exact path="/">
-            <List />
+            <ListView listId={listId} />
```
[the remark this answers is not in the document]
anderswift: Not sure if we mentioned this anywhere, but we tacked "View" onto the names of the components in /pages because it was getting confusing with the new components we were adding (as I'm sure you can imagine)!

src/components/AddItemForm/AddItemForm.js
```diff
@@ -0,0 +1,115 @@
...
+        type="text"
+        id="itemName"
+        name="itemName"
+        aria-describedby="itemNameHint"
+        value={formValues.itemName}
+        onChange={handleChange}
+        maxLength="100"
+        required
+      />
+      <span id="itemNameHint"></span>
```
abowler2, line 54: I agree with @Cjstoney's comment above regarding creating an issue for changes that you would like to implement in the future as opposed to leaving partial elements in the code. Particularly if they are not part of the AC.  You can mock out what you would like the code to be and include that in the issue description so that you have the needed reference to follow your original thought on it.

While the information you left here explaining your thoughts is helpful, on a bigger team there will be devs who will come upon this code and not have been involved with the original code review. In those cases they will not have the context to understand the intention behind this span.

Also, it is not uncommon in a project to meet the AC and continue moving forward in which case it might be weeks/months/years before this particular refinement is addressed (I'm currently discussing with my team a comment in our code base that was left to note a portion of code that should 'eventually' be refactored out and the comment is 3+ years old 😁 )

src/components/AddItemForm/AddItemForm.js
anderswift: The reason we used strings is due to input values being string type, so the quickest way we could think to get around that was to just use strings for the purchaseInterval values and comparisons, up until actually putting the value in the database. 

Any suggestions for a better way? I just tried out valueAsNumber, but it seems to return NaN for our radio buttons.

abowler2 (approved):
This is looking really good! 

I did want to point out that it looks like something happened when you updated as the `package-lock.json` added a large amount of code changes. Upon inspecting it locally it looks like many, if not all, packages were duplicated as `node-modules/.....`  

When I run `npm update` those additional lines of code are removed.

src/components/AddItemForm/AddItemForm.js
anderswift: Thanks for that helpful insight! I definitely agree that makes sense. Though TBH I didn't realize we'd left that span in there until I reviewed my own PR the next day. 😅

> I did want to point out that it looks like something happened when you updated as the `package-lock.json` added a large amount of code changes. Upon inspecting it locally it looks like many, if not all, packages were duplicated as `node-modules/.....`

Any idea what might have happened there? @connietran-dev mentioned something about a change in one of the packages files on her computer when we started, I think after running npm update (is that right Connie?). I assumed it was just a minor package update and wasn't paying a lot of attention to those files, whoops. Seems to be fixed on my end after running npm update as you suggested, and I pushed the fix.

abowler2: >Any idea what might have happened there?

I didn't when I saw it last night while reviewing, but I dug in a bit today and it is because @connietran-dev is using a newer version of `npm`.  The [updated v7](https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json) made significant changes to the lockfile for performance reasons but needed to stay backward compatible so they kept all the previous settings as well.  This explains why the information appeared to be in there twice.

I found this [blog/video](https://nitayneeman.com/posts/catching-up-with-package-lockfile-changes-in-npm-v7/) to be helpful in breaking down the changes made in npm v7.

We can discuss this at the sync tomorrow and decide on which version we should all use so that we don't end up with this file changing every week depending on who ran an update.

src/App.js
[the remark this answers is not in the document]
anderswift: @nick-zanetti I played around to kind of remind myself what JavaScript will and won't let you get away with, and if you were to do something like `someVar[0]`, `someVar.id` or `someVar[0].id` on an undefined or null value, you get TypeErrors and the app crashes. So I think I'm going to quickly add in that extra hasOwnProperty check as well!

src/App.js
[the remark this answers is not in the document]
yenly: These are really great ways to check. Sometimes libraries have a method for you to use so then your code would be easier to read. This is one of those use cases. Not a necessary change. Based on the [querySnapshot documentations](https://firebase.google.com/docs/reference/node/firebase.firestore.QuerySnapshot) @nick-zanetti  shared, you can use:
```javascript
if(!querySnaphort.empty) {
// save the list ide for later
}
```

yenly (approved):
Awesome work! 🚀  I like the way you built the form with accessibility in mind. I echo others by saying these component changes are forward-thinking and setting up for the next steps. Great idea in saving listId to state.

src/components/AddItemForm/AddItemForm.js
```diff
@@ -0,0 +1,113 @@
...
+  const handleSubmit = async (event) => {
+    event.preventDefault();
+
+    const newItem = formValues;
+    newItem.purchaseInterval = parseInt(newItem.purchaseInterval); // convert purchaseInterval to integer
+    newItem.lastPurchaseDate = null;
+    newItem.numberOfPurchases = 0;
+
+    try {
+      await db.collection(`lists/${listId}/items`).add(formValues); // add item to Firestore database
```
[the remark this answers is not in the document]
anderswift: Yes, we definitely did mean to add newItem. 🤦 Good catch because I completely missed that in the hustle to get things done.

PR #27, opened 2021-07-28
Join list form

### LINKED ISSUE #5: 5. As a user, I want to join an existing shopping list so I can share a shopping list with another person.
Tokens can be shared with other users to allow them to co-manage a given list.

Joining an existing list consists of the following:

* User enters an existing token
* Save the token to `localStorage`
* Show the user the existing list

AC:

* When the user does not already have a token in `localStorage`, on the onboarding/home screen, a simple form is displayed that allows the user to enter a token
* Entering the token and hitting submit saves the token to `localStorage`, effectively giving them joint control of the list
* On submit, show an error if the token does not exist
* If they get an error message, allow them to try again or create a new list

Notes:

* As of this story, the welcome/home screen will be feature complete, as reflected in this wireframe: https://github.com/the-collab-lab/smart-shopping-list/blob/main/_resources/wireframes/welcome-screen.png
* For accessibility reasons, use a `<label>` tag for the share token field.
* You can assume that users will only be sharing a token if the list has actual items on it. So, let's say I create a New Shopping List, but never add any items to it, and haven't saved that particular token to Firestore, we can assume that's a list that doesn't exist.
* If you would like, you can save the token to Firestore immediately when creating a list, and then a user would be able to share an empty list.

anderswift:
## Description

In this PR, we add functionality that allows a user to join a list. We also updated the `createList()` functionality to make the addItemForm and ShoppingList more compatible with the "create list" button functionality. 

#### Is Token Valid Function

`isTokenValid()` queries the database and checks for a match on the token parameter in the lists collection. 

#### Create List Function 

`createList()` generates a new token and then adds a list to the database with that token. It then saves the token in local storage if it was successful.

#### Join List Function

`joinList()` takes a user-provided token, checks that token with `isTokenValid()`, if it is, sets it to local storage. 

### In Home.js:

We added JSX for form with a controlled text input for joining a list and an aria-live region for errors. We then created two event handlers, `handleCreateList()` and `handleJoinList()`. `handleClick()` from the last sprint was changed to `handleCreateList()` to better differentiate between the two forms. 

### Styles

We added some basic CSS to make the working app look more like the wireframes. 

## Related Issue

Closes #5 

## Acceptance Criteria

AC:

- [x] When the user does not already have a token in localStorage, on the onboarding/home screen, a simple form is displayed that allows the user to enter a token
- [x] Entering the token and hitting submit saves the token to localStorage, effectively giving them joint control of the list
- [x] On submit, show an error if the token does not exist
- [x] If they get an error message, allow them to try again or create a new list

## Type of Changes

### Before

Screen Shot 2021-07-20 at 8 22 24 PM

### After

image
image

1) Run `npm install` (Reminder **not** to run `npm update`).
2) Make sure your local storage does not have an existing token, so that you are redirected to the home page.
3) Enter an invalid token (e.g. "not real token") to trigger the invalid token error message.
4) Enter a token for an existing list (e.g. "rodeo rage dip") to join that list. You should be redirected to the List view and have the token written to your localStorage.

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -40,6 +40,14 @@ BEM conventions we are using:
   width: 100%;
   font-size: 20px;
   margin: 5px 0 20px 0;
+  border: 1px solid #888;
+  border-radius: 3px;
+  padding: 8px 10px;
+}
+
+.text-field_has-error {
+  margin-bottom: 5px;
+  border: 2px solid red;
 }
 
 .button {
@@ -52,4 +60,10 @@ BEM conventions we are using:
 
 .error {
   color: red;
+  margin: 10px 0;
+}
+
+.error_type_field {
+  font-size: 16px;
+  margin: 0 0 20px 0;
 }
--- a/src/App.css
+++ b/src/App.css
@@ -18,5 +18,38 @@ BEM conventions we are using:
 */
 
 .container {
+  max-width: 500px;
+  margin: 0 auto;
+  font-size: 20px;
+  padding: 40px;
+  position: relative;
+}
+
+.container__header {
+  margin-bottom: 40px;
+}
+
+.container__separator {
   text-align: center;
+  margin: 40px 0;
+}
+
+.text-field {
+  display: block;
+  box-sizing: border-box;
+  width: 100%;
+  font-size: 20px;
+  margin: 5px 0 20px 0;
+}
+
+.button {
+  font-size: 20px;
+  width: 100%;
+  box-sizing: border-box;
+  padding: 8px;
+  margin: 10px 0;
+}
+
+.error {
+  color: red;
 }
--- a/src/App.js
+++ b/src/App.js
@@ -30,9 +30,9 @@ function App() {
 
     return db
       .collection('lists')
-      .add({ token: token })
+      .add({ token: token }) //add a list and assign it the user's token
       .then((results) => {
-        const { id } = results;
+        const { id } = results; //get the firestore-generated id from the created list
         if (id === undefined) throw new Error('Failed to create list.');
         return id;
       })
--- a/src/App.js
+++ b/src/App.js
@@ -24,13 +24,24 @@ import getToken from './lib/tokens';
 
 function App() {
   const [listId, setListId] = useState(null);
-  const [userToken, setUserToken] = useState('');
 
-  function saveToken(e) {
-    e.preventDefault();
+  function createList() {
     const token = getToken();
-    localStorage.setItem('token', token);
-    setUserToken(token);
+
+    return db
+      .collection('lists')
+      .add({ token: token })
+      .then((results) => {
+        const { id } = results;
+        if (id === undefined) throw new Error('Failed to create list.');
+        return id;
+      })
+      .then((newListId) => {
+        // list successfully created in db, so update listId state and save token to localStorage
+        setListId(newListId);
+        localStorage.setItem('token', token);
+        return newListId;
+      });
   }
 
   // on component mounting, look for token in local storage and use it to retrieve the list id
@@ -38,8 +49,6 @@ function App() {
     const token = localStorage.getItem('token');
 
     if (token) {
-      setUserToken(token);
-
       // find the list in Firestore associated with the stored token from local storage
       db.collection('lists')
         .where('token', '==', token)
@@ -54,32 +63,24 @@ function App() {
           console.log('Error getting list: ', error);
         });
     }
-  }, [listId]);
+  }, []);
 
   return (
     <Router>
       <div className="App container">
         <Switch>
           <Route exact path="/">
-            {userToken ? (
+            {listId ? (
               <Redirect to="/list" />
             ) : (
-              <Home
-                userToken={userToken}
-                setUserToken={setUserToken}
-                saveToken={saveToken}
-              />
+              <Home createList={createList} />
             )}
           </Route>
           <Route path="/list">
-            {!userToken ? (
-              <Redirect exact to="/" />
-            ) : (
-              <ListView listId={listId} />
-            )}
+            {!listId ? <Redirect exact to="/" /> : <ListView listId={listId} />}
           </Route>
           <Route path="/add">
-            {!userToken ? (
+            {!listId ? (
               <Redirect exact to="/" />
             ) : (
               <AddItemView listId={listId} />
--- a/src/App.js
+++ b/src/App.js
@@ -40,7 +40,26 @@ function App() {
         // list successfully created in db, so update listId state and save token to localStorage
         setListId(newListId);
         localStorage.setItem('token', token);
-        return newListId;
+        return true;
+      });
+  }
+
+  function joinList(shareToken) {
+    return db
+      .collection('lists')
+      .where('token', '==', shareToken)
+      .get()
+      .then((querySnapshot) => {
+        // if there are results and an id property exists
+        if (!querySnapshot.empty && 'id' in querySnapshot.docs[0]) {
+          setListId(querySnapshot.docs[0].id);
+          localStorage.setItem('token', shareToken);
+          return true;
+
+          // otherwise, the token is invalid
+        } else {
+          throw new Error('Token is invalid.');
+        }
       });
   }
 
@@ -73,7 +92,7 @@ function App() {
             {listId ? (
               <Redirect to="/list" />
             ) : (
-              <Home createList={createList} />
+              <Home createList={createList} joinList={joinList} />
             )}
           </Route>
           <Route path="/list">
--- a/src/App.js
+++ b/src/App.js
@@ -81,7 +81,7 @@ function App() {
     if (token) {
       isTokenValid(token)
         .then((listExists) => {
-          if (listExists !== true) localStorage.removeItem('token');
+          if (!listExists) localStorage.removeItem('token');
         })
         .catch((error) => {
           console.log('Error getting list: ', error);
--- a/src/App.js
+++ b/src/App.js
@@ -28,20 +28,29 @@ function App() {
   function createList() {
     const token = getToken();
 
-    return db
-      .collection('lists')
-      .add({ token: token }) //add a list and assign it the user's token
-      .then((results) => {
-        const { id } = results; //get the firestore-generated id from the created list
-        if (id === undefined) throw new Error('Failed to create list.');
-        return id;
-      })
-      .then((newListId) => {
-        // list successfully created in db, so update listId state and save token to localStorage
-        setListId(newListId);
-        localStorage.setItem('token', token);
-        return true;
-      });
+    // before attempting to add new list, make sure token hasn't been used before in the db
+    return isTokenValid(token).then((listExists) => {
+      if (listExists) {
+        // if isTokenValid returns true the token is taken, so try again
+        createList();
+      } else {
+        // otherwise the token is not already associated with a list, so we can safely use it
+        return db
+          .collection('lists')
+          .add({ token: token }) // add a list and assign it the user's token
+          .then((results) => {
+            const { id } = results; // get the firestore-generated id from the created list
+            if (id === undefined) throw new Error('Failed to create list.');
+            return id;
+          })
+          .then((newListId) => {
+            // list successfully created in db, so update listId state and save token to localStorage
+            setListId(newListId);
+            localStorage.setItem('token', token);
+            return true;
+          });
+      }
+    });
   }
 
   // returns true if token matches a list in db, returns false if no list, throws error on connection problem
--- a/src/App.js
+++ b/src/App.js
@@ -44,23 +44,34 @@ function App() {
       });
   }
 
-  function joinList(shareToken) {
+  // returns true if token matches a list in db, returns false if no list, throws error on connection problem
+  function isTokenValid(token) {
     return db
       .collection('lists')
-      .where('token', '==', shareToken)
+      .where('token', '==', token)
       .get()
       .then((querySnapshot) => {
         // if there are results and an id property exists
         if (!querySnapshot.empty && 'id' in querySnapshot.docs[0]) {
           setListId(querySnapshot.docs[0].id);
-          localStorage.setItem('token', shareToken);
           return true;
 
-          // otherwise, the token is invalid
+          // check metadata.fromCache to distinguish between no results (invalid token) and a connection issue
+        } else if (!querySnapshot.metadata.fromCache) {
+          return false;
         } else {
-          throw new Error('Token is invalid.');
+          throw new Error('Connection problem');
         }
       });
+  }
+
+  function joinList(shareToken) {
+    return isTokenValid(shareToken).then((listExists) => {
+      if (listExists) {
+        localStorage.setItem('token', shareToken);
+        return true;
+      } else throw new Error('Invalid token');
+    });
   }
 
   // on component mounting, look for token in local storage and use it to retrieve the list id
@@ -68,15 +79,9 @@ function App() {
     const token = localStorage.getItem('token');
 
     if (token) {
-      // find the list in Firestore associated with the stored token from local storage
-      db.collection('lists')
-        .where('token', '==', token)
-        .get()
-        .then((querySnapshot) => {
-          // if there are results and an id property exists
-          if (!querySnapshot.empty && 'id' in querySnapshot.docs[0]) {
-            setListId(querySnapshot.docs[0].id); // save the list id for later
-          }
+      isTokenValid(token)
+        .then((listExists) => {
+          if (listExists !== true) localStorage.removeItem('token');
         })
         .catch((error) => {
           console.log('Error getting list: ', error);
--- a/src/components/Header/Header.css
+++ b/src/components/Header/Header.css
@@ -1,0 +1,13 @@
+.header {
+  text-align: center;
+}
+
+.header__welcome {
+  margin: 0;
+  font-size: 36px;
+}
+
+.header__logo {
+  margin: 0;
+  font-size: 30px;
+}
--- a/src/components/NavMenu/NavMenu.js
+++ b/src/components/NavMenu/NavMenu.js
@@ -3,12 +3,12 @@ import './NavMenu.css';
 
 const NavMenu = () => {
   return (
-    <nav className="container__nav-menu nav-menu" aria-label="Main Navigation">
+    <nav className="container__nav-menu nav-menu" aria-label="Main Menu">
       <ul className="nav-menu__list">
         <li>
           <NavLink
             exact
-            to="/"
+            to="/list"
             className="nav-menu__item"
             activeClassName="nav-menu__item_current"
           >
--- a/src/pages/Home/Home.js
+++ b/src/pages/Home/Home.js
@@ -108,7 +108,7 @@ function Home({ createList, joinList }) {
           />
           <div
             id="shareTokenHint"
-            class={`error error_type_field ${
+            className={`error error_type_field ${
               shareTokenError ? 'error_on' : ''
             }`}
           >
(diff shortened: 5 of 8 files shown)


src/App.js
```diff
@@ -24,62 +24,87 @@ import getToken from './lib/tokens';
...
+        const { id } = results;
+        if (id === undefined) throw new Error('Failed to create list.');
+        return id;
+      })
+      .then((newListId) => {
+        // list successfully created in db, so update listId state and save token to localStorage
+        setListId(newListId);
+        localStorage.setItem('token', token);
+        return true;
+      });
```
anderswift, line 44: I think it would be good practice to add a check in here, to make sure the token doesn't already exist in the database before adding it as a unique identifier for an account. I think we could do it pretty easily with our new isTokenValid function.
```suggestion
    return isTokenValid(token)
      .then((listExists) => {
        if(listExists) {
          createList(); // this token is taken, so try again
        } else {
          return db
          .collection('lists')
          .add({ token: token })
          .then((results) => {
            const { id } = results;
            if (id === undefined) throw new Error('Failed to create list.');
            return id;
          })
          .then((newListId) => {
            // list successfully created in db, so update listId state and save token to localStorage
            setListId(newListId);
            localStorage.setItem('token', token);
            return true;
          });
        }
      });
```

src/App.js
```diff
@@ -24,62 +24,87 @@ import getToken from './lib/tokens';
...
+      .collection('lists')
+      .add({ token: token })
+      .then((results) => {
+        const { id } = results;
+        if (id === undefined) throw new Error('Failed to create list.');
+        return id;
+      })
+      .then((newListId) => {
+        // list successfully created in db, so update listId state and save token to localStorage
+        setListId(newListId);
```
connietran-dev, line 41: For my own understanding, I assume you are de-structuring the list's id out of the results from Firestore, then using that as newListId and setting that as listId? A few comments here might have been helpful for me personally!

src/pages/Home/Home.js
```diff
@@ -1,24 +1,102 @@
...
+      })
+      .catch((err) => {
+        console.log(err);
+      });
+  }
+
+  function handleJoinList(event) {
+    event.preventDefault();
+    setJoinListError(''); // remove message to ensure repeated error is read again by screen reader
+    joinList(shareToken)
```
connietran-dev, line 28: Where does  `shareToken` come from that is passed into `joinList` here? It is the user-provided token? For my understanding, where are you getting the token from in this function?

connietran-dev: Are we essentially passing in the user's token into `joinList`? I'm not sure I intuitively understood the variable name `shareToken` at first. Can we just call it userToken? The more I look at it, the more it makes sense to me as the token used in order to share...but just my initial thoughts!

src/App.js
```diff
@@ -24,62 +24,87 @@ import getToken from './lib/tokens';
...
-            {userToken ? (
+            {listId ? (
               <Redirect to="/list" />
             ) : (
-              <Home
-                userToken={userToken}
-                setUserToken={setUserToken}
-                saveToken={saveToken}
-              />
+              <Home createList={createList} joinList={joinList} />
```
connietran-dev, line 100: Is there a reason why we can't have the `createList` and `joinList` functions in Home.js instead of passing them down as props in App.js? 

There may be an architectural or best practice reason in React, but it was hard for me to jump between Home.js and App.js to understand what the functions were doing.

src/App.js
```diff
@@ -24,62 +24,87 @@ import getToken from './lib/tokens';
...
 
   return (
     <Router>
       <div className="App container">
         <Switch>
           <Route exact path="/">
-            {userToken ? (
+            {listId ? (
               <Redirect to="/list" />
             ) : (
```
connietran-dev, line 99: Is there a reason why we have changed from checking userToken in the routes to listId?

src/App.js
```diff
@@ -24,62 +24,87 @@ import getToken from './lib/tokens';
...
-        .where('token', '==', token)
-        .get()
-        .then((querySnapshot) => {
-          // if there are results and an id property exists
-          if (!querySnapshot.empty && 'id' in querySnapshot.docs[0]) {
-            setListId(querySnapshot.docs[0].id); // save the list id for later
-          }
+      isTokenValid(token)
+        .then((listExists) => {
+          if (listExists !== true) localStorage.removeItem('token');
```
connietran-dev, line 84: When our component mounts, we are setting token to what is in local storage. Then we are checking `isTokenValid` against the database. If the `listExists` does not exist in the db, and thus is not equal to true, aka it is false, we are removing the token from local storage correct?

If this happened, we're assuming the user has an invalid token in local storage for some reason, so we are removing it, and the user would see the Home screen where they can create a new list/token or join an existing list/token.

I want to make sure my understanding is correct!

src/App.css
```diff
@@ -18,5 +18,38 @@ BEM conventions we are using:
...
+}
+
+.container__separator {
   text-align: center;
+  margin: 40px 0;
+}
+
+.text-field {
+  display: block;
+  box-sizing: border-box;
```
connietran-dev, line 39: We've put border-box box-sizing on this particular class, but do we want to set it globally for all elements?

src/components/Header/Header.css
```diff
@@ -0,0 +1,13 @@
+.header {
+  text-align: center;
+}
+
+.header__welcome {
+  margin: 0;
```
connietran-dev, line 6: Do we have to have a global reset with margin and padding 0?

src/components/NavMenu/NavMenu.js
```diff
@@ -3,12 +3,12 @@ import './NavMenu.css';
 
 const NavMenu = () => {
   return (
-    <nav className="container__nav-menu nav-menu" aria-label="Main Navigation">
+    <nav className="container__nav-menu nav-menu" aria-label="Main Menu">
```
connietran-dev, line 6: I agree with this change! It's clearer!

connietran-dev:
Great job on this functionality! 

There is definitely a lot of great checking of the database and logic that you all have built. The styling of the CSS looks really crisp as well.

Most of my comments are just for my own understanding of the code and how it all works together. If it's easier, we can just handle it offline or in a call, but it may help with @SheilaRMoore 's review of the code as well. And if there is somewhere we can add comments to help clarify functionality, that will probably help with the long-term. 

I did notice one thing, and not sure which line is causing this, but if you already have a token in local storage, there is some flashing of the Home screen before showing the List. Not sure if this is intentional with the different redirects we have.

Otherwise, great job all!

connietran-dev (approved):

src/pages/Home/Home.js
anderswift: `shareToken` is the value of the form field labelled "Share Token" in the wireframe, and with the name and id of "shareToken"—basically we used the exact same pattern of a controlled form field component as in the "Add Item" form, but since there will only ever be one input we named the input's linked state variable consistent with the input. Given that, I think shareToken (or maybe shareTokenValue) makes sense, but if you still feel like userToken is clearer I think we should change it in the id/name/classes as well to be consistent.

src/components/NavMenu/NavMenu.js
anderswift: Mostly listening to the screen reader say "Main Navigation Navigation" was driving me nuts! 😂

src/components/Header/Header.css
anderswift: Do you mean like css reset or normalize.css? We could do that, though I think they're kind of falling out of vogue with the idea that it's much less overhead to not have redundant rules, rules for elements you aren't even using etc. and the resets all come with their own different defaults to learn anyway. Since this is a fairly small project I think I favour just being explicit about margins, padding and font sizes in the BEM classes, because then it takes the guess work out for another developer looking at the styles.

src/App.js
anderswift: Yes, we check localStorage for an existing token when the app first loads. Our function isTokenValid checks the database for the token, and also accounts for the possibility that an empty result might be because there is no list for that token (therefore it is invalid, the function returns false, and we delete the bad token) or might be because the person is having a connection issue, in which case we leave things alone and just print an error to the console. Line 47 comment gives a quick overview of the isTokenValid function and possible return values and what they indicate.

src/App.js
anderswift: Yep! Do you remember when we built the "Add Item" form, and I mentioned we needed to add a check somewhere to make sure we had a list ID before loading it? Since the item form and shopping list won't work without a list ID (won't know which list to read/write to), we switched the routing logic to check for a list ID instead. You may have noticed we also create a new list in the database when a new token is generated, so a token will always have a list if it is valid.

src/App.js
anderswift: Good suggestion! I've made a note to add some in.

src/App.css
anderswift: I'd worry that might be a bit counterintuitive/confusing, since a person wouldn't necessarily expect that behaviour for all elements. I just did this on a few form elements to easily make them 100% width without having to account for padding, border, etc.

src/App.js
anderswift: Yes, good observation! This is mostly to do with the way useHistory works, and partly because it made sense to maintain error states and values for the Home view forms inside the Home component. 

It may help to think of it like this: 99% of the work for creating a list or joining a list is handled by the functions declared in App.js. Inside the Home component, we want to use these functions when our "Create List" button is pushed, or when our "Join List" form is submitted, but we also need to do a couple of "local" things in Home based on the results: 
1. We need to handle redirects in Home, because useHistory won't work properly in App.js (I believe @nick-zanetti has a link with more info on this behaviour⁠—Nick, can you share?)
2. We need to update the error state for the Join List form to show error messages to the user, and seemed to make the most sense to keep this state in the Home component.

We added `handleCreateList` and `handleJoinList` functions to handle events in Home, and they simply run the `createList`/`joinList` functions then do the extra bits above.

src/pages/Home/Home.js
```diff
@@ -1,24 +1,102 @@
...
+          <input
+            className="join-list-form__text-field text-field"
+            type="text"
+            id="shareToken"
+            name="shareToken"
+            value={shareToken}
+            onChange={handleTokenChange}
+            maxLength="100"
+            required
+          />
```
anderswift, line 92: A couple things I think we may want to add for a11y reasons:

1. Add `aria-invalid="true"` to the input when appropriate to indicate when that form field is associated with an error
2. Since we will only have one field that can have errors, adding `aria-describedBy` on the input + and a matching id on the aria live region might be helpful to tie them together for a user focused on the field (though I want to test this more with the next suggestion to make sure it doesn't trigger too many screen reader announcements)
3. Adding a ref and focusing on the field when there is an error would be better UX, especially for disabled users

src/pages/Home/Home.js
nick-zanetti: Yes, `shareToken` is the token the user enters in the bottom form field on the Home screen to join a list. And yes, that user provided token is passed to `joinList`. The "share token" name comes from the [wireframe](https://github.com/the-collab-lab/smart-shopping-list/blob/main/_resources/wireframes/welcome-screen.png) provided for this story, as Ander said. Given that, I think I would err on the side of keeping it as `shareToken`, because I assume future project stories might refer to it as that?

src/App.js
nick-zanetti: So to expand on what Ander was saying, a problem Sheila and I ran into last week was that `useHistory()` must be called in a _child functional component_. The [history instance](https://reactrouter.com/web/api/history) (which `useHistory` gives us access to) uses a context that is set by the `<Router>` component (in our case, I think this context pretty much means the URL slugs/routes). If we call `useHistory()` in App.js, that context will not be set in time for us to use it, so it will be undefined. More on this [here](https://stackoverflow.com/questions/62614433/react-router-why-is-the-usehistory-undefined-in-react). I'm still trying to wrap my head around this so hopefully I didn't butcher that explanation 😅

src/App.js
nick-zanetti: To put it another way, the `listId` is also effectively a check for userToken, because our app will not set a `listId` state without a token.

src/App.js
SheilaRMoore: I agree with Connie as far as adding some comments. I'm also thinking along the lines of what if this was a work situation. Where maybe someone would have to come behind you blindly without knowing the AC or the codebase already. They maybe a little confused here. However, I do really think that the if statement is pretty slick because of the "throw". I never really see it that often, and I feel like you probably eliminated several lines of code just by using that. Really good job here.

src/pages/Home/Home.js
SheilaRMoore: Looking at Nick's comment and looking at the next few stories that we have coming up, I think it is ok to keep it as shareToken or shareTokenValue, since now we have removed userToken an replaced with createList and joinList.

src/App.js
SheilaRMoore: Yes I remember this issue that Nick and I had. After some research, I believe we found where it would only work in child components, and that it won't work when it's within the same components where the Routes are listed.

src/App.css
SheilaRMoore: I think it makes sense for now to leave it at 100%. Looking ahead and the user stories coming up I think this may change. However, I think it is ok for now.

src/components/NavMenu/NavMenu.js
SheilaRMoore: Yeah it would probably drive me nuts, too. Lol. But I am glad you made the change here.

src/pages/Home/Home.js
SheilaRMoore: Now that I'm really learning a11y, I have to agree with you here. I also wanted to thank you and Connie for teaching me about that. I  think the aria-invalid="true" would be useful. Overall I think you guys did a great job, and that the only suggestion I have is to maybe add a few more comments. If you do add it , let me know. I would definitely like to see it in action with the screen reader.

SheilaRMoore (approved):

src/pages/Home/Home.js
anderswift: After thinking about it and lots of testing different arrangements, I think it actually makes sense to have 2 "error regions" here: 

1. a hint for the input (the aria-describedby on the input refers to the id of this area), and
2. a summary of the new state of the entire form (aria-live or role=alert).

The first works in combination with aria-invalid=true when there is something wrong with the specific field (e.g. invalid token). This is really important, because it ensures any time the user focuses on the field, they will be told it has an error and read the describedby hint, which could be as simple as "Invalid token." 

The second I think would work better for our "connection problem" message, since it really has nothing to do with the specific field so would be misleading to connect it as describedby and make the field invalid. This area can also be used for a more detailed message for an invalid token, because it only gets triggered once (ideally right after form submission if the screen reader is being predictable). So something like "Sorry, your token is invalid. Please try again or create a new list" could go here. The images below should give you a better idea of what I mean. (If it seems too redundant to show both, we could always visually hide one message when we add final CSS.) 

image

image

src/App.js
anderswift: Actually @nick-zanetti, I think the way you put it makes perfect sense! Very good explanation.

src/pages/Home/Home.js
anderswift: Just stuck in some comments to further clarify the state variables we are using, since I added another one for a11y updates. I think it should be clearer now.

src/App.js
```diff
@@ -24,62 +24,87 @@ import getToken from './lib/tokens';
...
```
yenly, line 84: This is not a necessary change and just sharing a common pattern:
```javascript
if (!listExists) localStorage.removeItem('token');
```

src/pages/Home/Home.js
```diff
@@ -1,24 +1,126 @@
...
+            'Sorry, there was a problem with your token. Please try again or create a new list.',
+          );
+          shareTokenRef.current.focus();
+        } else {
+          setJoinListError(
+            'Sorry, there was a problem connecting to the database. Please try again.',
+          );
+          shareTokenRef.current.focus();
+        }
+      });
```
yenly, line 52: I like the useRef approach here to put the focus on the input field to encourage the user to enter again. 🥇 for usability.

yenly (approved):
👏🏽 Amazing work! Love the thoughtfulness in handling usability and accessibility.

src/App.js
anderswift: IIRC this is an artifact of at an earlier point explicitly checking for boolean false vs. JS "falsy", but as the code is now I think your suggestion would read clearer.

abowler2 (approved):
Really nice work. I like that you don't just give an error but you also let the user know what the problem was that caused the error.

src/App.css
anderswift: Yeah, these are just barebones styles to make the app less painful to look at while building, hence trying to do the minimum. I expect we'll revisit and better optimize the CSS once we finalize a design.

src/pages/Home/Home.js
```diff
@@ -1,24 +1,126 @@
...
-    <main>
-      <h1>Welcome To Your Smart Shopping List</h1>
-      <button type="button" onClick={handleClick}>
-        Create a new list
-      </button>
-    </main>
+    <>
+      <header className="container__header header">
+        <h1 className="header__welcome">Welcome To Your Smart Shopping List</h1>
+      </header>
```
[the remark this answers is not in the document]
anderswift: @abowler2 Yes, I had the same thought! Then considered we would very likely end up with something entirely different by the design/branding stage, so wanted to give that more thought. (For example, I'd guess we might want a large logo and some kind of introduction/welcome/tagline on Home, then a smaller logo on other views; logos should share a block class with a home modifier... I actually think I know good a way to do it now so will come back to this!)

src/App.js
```diff
@@ -24,62 +24,87 @@ import getToken from './lib/tokens';
...
+        const { id } = results; //get the firestore-generated id from the created list
+        if (id === undefined) throw new Error('Failed to create list.');
+        return id;
+      })
+      .then((newListId) => {
+        // list successfully created in db, so update listId state and save token to localStorage
+        setListId(newListId);
+        localStorage.setItem('token', token);
+        return true;
+      });
```
anderswift, line 44: Updating my suggestion to check for token conflicts in the database first (since suggested code is outdated with the new comments):
```suggestion
    // before attempting to add new list, make sure token hasn't been used before in the db  
    return isTokenValid(token)   
      .then((listExists) => {
        if(listExists) { 
          // if isTokenValid returns true the token is taken, so try again
          createList();
        } else {
          // otherwise the token is not already associated with a list, so we can safely use it
          return db
          .collection('lists')
          .add({ token: token }) // add a list and assign it the user's token
          .then((results) => {
            const { id } = results; // get the firestore-generated id from the created list
            if (id === undefined) throw new Error('Failed to create list.');
            return id;
          })
          .then((newListId) => {
            // list successfully created in db, so update listId state and save token to localStorage
            setListId(newListId);
            localStorage.setItem('token', token);
            return true;
          });
        }
      });
```

src/App.js
```diff
@@ -24,62 +24,87 @@ import getToken from './lib/tokens';
...
+        // if there are results and an id property exists
+        if (!querySnapshot.empty && 'id' in querySnapshot.docs[0]) {
+          setListId(querySnapshot.docs[0].id);
+          return true;
+
+          // check metadata.fromCache to distinguish between no results (invalid token) and a connection issue
+        } else if (!querySnapshot.metadata.fromCache) {
+          return false;
+        } else {
+          throw new Error('Connection problem');
```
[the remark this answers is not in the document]
Cjstoney: +1 To Yenly's comment. I am super impressed with the ability to read docs and more importantly understand the docs

Cjstoney (approved):
I think this PR looks and functions extremely well. As stated before, I am impressed with how well you all are consuming and then applying info from the documentation. That is a tough skill to acquire.

PR #28, opened 2021-07-28
Prevent duplicate item

### LINKED ISSUE #6: 6. As a user, I want to be alerted when I’m entering an item that is the same as something already on my list so I can avoid duplicate items.
AC:

* Show an error message if the user tries to submit a new item that has the exact same name as an existing item
* Show an error message if the user tries to submit a new item that has the same name as an existing item, where capitalization has been normalized and punctuation has been removed
* The user’s original input should be what gets saved in the database

SheilaRMoore:
## Description

When a user tries to add a new item, the AddItemForm retrieves the itemNames out of Firestore and compares against the itemName of the new item being entered. If they match, an error message displays under Item Name that this item already exists in the Shopping List.

Additional Notes: 
- We decided to prevent the user from, say, adding "jackfruit" and "jack fruit" by removing spaces. You can STILL add spaces and punctuation when you FIRST add an item to the database. However, when you try to add "jackfruit" AGAIN, it will remove spaces, which will prevent you from adding "jack fruit".
- We also decided to remove any spaces before and after what the user enters from being entered into the database in order to prevent issues in the database. We followed the Microsoft To-Do app in this practice. This also makes it easier to compare itemNames when checking for duplicates.
- We also did some grouping of variables/state/functions in the AddItemForm component. I think this is a practice we have all been following, but we codified it with comments.

## Related Issue

Closes #6: As a user, I want to be alerted when I’m entering an item that is the same as something already on my list so I can avoid duplicate items

## Acceptance Criteria

- Show an error message if the user tries to submit a new item that has the exact same name as an existing item
- Show an error message if the user tries to submit a new item that has the same name as an existing item, where 
   capitalization has been normalized and punctuation has been removed
- The user’s original input should be what gets saved in the database

## Type of Changes

## Updates

In the AddItemForm component, we added a new span under the Item Name input to display the duplicate error message. 

In the handleSubmit function, when the user hits Add Item, it runs the following new functionality:

- Gets array of listItems from Firestore (db.collection.get().then(querySnapshot))
- Checks new function - isItemDuplicate: Checks if the itemName being entered matches any of the itemName's from Firestore
- Which uses normalizeInput: Removes punctuation and spaces from the itemName being entered and does the same for the array of items out of Firestore
- If isItemDuplicate is true, then 'Item already exists in Shopping List' is setErrorMessage.
- If isItemDuplicate is false, then we addItemToDatabase - which is previous functionality that we moved into its own function

### Before

image

### After

image

1. Pull down branch.
2. npm start
3. Add an item that does not yet exist in list.
4. Try adding an item that has the exact same name. The error will appear.
5. Try adding an item that has the same name but with punctuation in between the characters. (ex: honey and h@#$oney). The error will appear.
6. Try adding an item that has the same name but with spaces in between the characters (ex: jack fruit and jackfruit). The error will appear.

## To-Dos / Nice to Haves
Some things we'd like to do but didn't get a chance to get to:

- We noticed Issue #29 
- Make setErrorMessage for displaying the error message more generic so that if we were ever to add more input fields, it would still work.

### DIFF
--- a/src/components/AddItemForm/AddItemForm.js
+++ b/src/components/AddItemForm/AddItemForm.js
@@ -14,7 +14,14 @@ const AddItemForm = ({ listId }) => {
 
   const isItemDuplicate = (item, array) => {
     // TODO: Potential to remove all spaces in item and items in array
-    const itemToCompare = item.toLowerCase().trim(); // transform and remove leading and/or trailing whitespace
+
+    // .replace's remove punctuation, then remove extra spaces from removing punctuation
+    // .trim() to remove leading and/or trailing whitespace (must be last)
+    const itemToCompare = item
+      .toLowerCase()
+      .replace(/[.,\/#!$%^&*;:{}=\-_`~()@[\]|<>+'"/?]/g, '')
+      .replace(/\s{2,}/g, ' ')
+      .trim();
 
     // if .indexOf returns -1, the item does not exist in array, aka it's not a duplicate, aka false
     return array.indexOf(itemToCompare) === -1 ? false : true;
@@ -22,7 +29,7 @@ const AddItemForm = ({ listId }) => {
 
   const addItemToDatabase = async () => {
     const newItem = {
-      ...formValues,
+      itemName: formValues.itemName.trim(), // remove extra whitespace from itemName to keep data clean
       purchaseInterval: Number(formValues.purchaseInterval),
       lastPurchaseDate: null,
       numberOfPurchases: 0,
--- a/src/components/AddItemForm/AddItemForm.js
+++ b/src/components/AddItemForm/AddItemForm.js
@@ -7,10 +7,10 @@ const AddItemForm = ({ listId }) => {
     purchaseInterval: '7',
   };
 
-  const defaultInputMessage = '';
+  const defaultErrorMessage = '';
 
   const [formValues, setFormValues] = useState(defaultFormValues);
-  const [inputMessage, setInputMessage] = useState(defaultInputMessage);
+  const [errorMessage, setErrorMessage] = useState(defaultErrorMessage);
 
   // used when comparing entered item and array of db items for duplicates
   const normalizeInput = (item) => {
@@ -49,7 +49,7 @@ const AddItemForm = ({ listId }) => {
 
   // generic function updates formValues state for any of the below form inputs
   const handleChange = (event) => {
-    setInputMessage(defaultInputMessage); // if input field changes, reset error message
+    setErrorMessage(defaultErrorMessage); // if input field changes, reset error message
 
     const inputName = event.target.name;
     setFormValues({ ...formValues, [inputName]: event.target.value });
@@ -78,7 +78,7 @@ const AddItemForm = ({ listId }) => {
 
           // if item exists, show error message, otherwise, continue with adding to database
           if (duplicateResult === true) {
-            setInputMessage('Item already exists in Shopping List');
+            setErrorMessage('Item already exists in Shopping List');
           } else {
             addItemToDatabase();
           }
@@ -101,13 +101,13 @@ const AddItemForm = ({ listId }) => {
         type="text"
         id="itemName"
         name="itemName"
-        aria-describedby="itemNameMessage"
+        aria-describedby="itemErrorMessage"
         value={formValues.itemName}
         onChange={handleChange}
         maxLength="100"
         required
       />
-      <span id="itemNameMessage">{inputMessage}</span>
+      <span id="itemErrorMessage">{errorMessage}</span>
 
       <fieldset>
         <legend>How soon will you buy this again?</legend>
--- a/src/components/AddItemForm/AddItemForm.js
+++ b/src/components/AddItemForm/AddItemForm.js
@@ -58,8 +58,6 @@ const AddItemForm = ({ listId }) => {
   };
 
   const isItemDuplicate = (item, array) => {
-    // TODO: Potential to remove all spaces in item and items in array
-
     const itemToCompare = normalizeInput(item);
     const arrayToCompare = array.map((dbItem) => normalizeInput(dbItem));
 
@@ -71,8 +69,8 @@ const AddItemForm = ({ listId }) => {
   const normalizeInput = (item) => {
     return item
       .toLowerCase()
-      .replace(/[.,/#!$%^&*;:{}=\-_`~()@[\]|<>+'"/?]/g, '') // remove punctuation
-      .replace(/\s{2,}/g, ' ') // remove extra spaces from removing punctuation
+      .replace(/\W/g, '') // remove non-word characters aka remove punctuation
+      .replace(/ /g, '') // remove spaces
       .trim(); // remove leading and trailing whitespace (trim must be last)
   };
 
--- a/src/components/AddItemForm/AddItemForm.js
+++ b/src/components/AddItemForm/AddItemForm.js
@@ -35,18 +35,13 @@ const AddItemForm = ({ listId }) => {
         .collection(`lists/${listId}/items`)
         .get()
         .then(async (querySnapshot) => {
-          // create array (dbItemArray) by returning itemNames from Firestore response (querySnapshot)
-          let dbItemArray = querySnapshot.docs.map((doc) => {
-            return doc.data().itemName;
-          });
-
-          const duplicateResult = isItemDuplicate(
-            formValues.itemName,
-            dbItemArray,
+          // create array of normalized item names from Firestore response (querySnapshot)
+          const dbItemArray = querySnapshot.docs.map((doc) =>
+            normalizeInput(doc.data().itemName),
           );
 
           // if item exists, show error message, otherwise, continue with adding to database
-          if (duplicateResult === true) {
+          if (dbItemArray.includes(normalizeInput(formValues.itemName))) {
             setErrorMessage('Item already exists in Shopping List');
           } else {
             addItemToDatabase();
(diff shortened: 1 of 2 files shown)


src/components/AddItemForm/AddItemForm.js
```diff
@@ -19,8 +28,55 @@ const AddItemForm = ({ listId }) => {
   const handleSubmit = async (event) => {
     event.preventDefault();
```
anderswift, line 30: Something I noticed when testing my own PR's form: it's important to clear any old errors when the form is submitted for accessibility reasons.

If a user tries to submit the same bad input twice in a row, and you don't clear the error, there's no feedback for a visually-impaired user after the first time. Resetting the error message here would work really well with my later suggestion to add in an aria live region.
```suggestion
setErrorMessage('');

```

src/components/AddItemForm/AddItemForm.js
```diff
@@ -2,15 +2,24 @@ import { useState } from 'react';
 import { db } from '../../lib/firebase.js';
 
 const AddItemForm = ({ listId }) => {
+  /** Default Values **/
   const defaultFormValues = {
     itemName: '',
     purchaseInterval: '7',
   };
 
+  const defaultErrorMessage = '';
```
anderswift, line 11: I'm curious why the defaultErrorMessage variable if it's just for setting the error to an empty string? It was a little counterintuitive to me when reading further down in the code. For example when I see `setErrorMessage(defaultErrorMessage);`, I'm not immediately clear on what you are doing unless I refer back to the line where you declare the variable. Whereas if you just did `setErrorMessage('');` I'd know without looking up from that line exactly what was going on.

src/components/AddItemForm/AddItemForm.js
```diff
@@ -19,8 +28,55 @@ const AddItemForm = ({ listId }) => {
...
+      .toLowerCase()
+      .replace(/\W/g, '') // remove non-word characters aka remove punctuation
+      .replace(/ /g, '') // remove spaces
+      .trim(); // remove leading and trailing whitespace (trim must be last)
+  };
+
+  const addItemToDatabase = async () => {
     const newItem = {
-      ...formValues,
+      itemName: formValues.itemName.trim(), // remove extra whitespace from itemName to keep data clean
```
anderswift, line 79: Good idea!

src/components/AddItemForm/AddItemForm.js
```diff
@@ -19,8 +28,55 @@ const AddItemForm = ({ listId }) => {
...
+          });
+
+          const duplicateResult = isItemDuplicate(
+            formValues.itemName,
+            dbItemArray,
+          );
+
+          // if item exists, show error message, otherwise, continue with adding to database
+          if (duplicateResult === true) {
+            setErrorMessage('Item already exists in Shopping List');
```
anderswift, line 50: For accessibility, it would be nice to focus the user on the text field after submission if there is an error, to save having to navigate to the invalid field. (We did not do this in the other PR's form either, so I will make a suggestion there as well.)

src/components/AddItemForm/AddItemForm.js
```diff
@@ -47,11 +103,13 @@ const AddItemForm = ({ listId }) => {
         type="text"
         id="itemName"
         name="itemName"
+        aria-describedby="itemErrorMessage"
```
anderswift, line 106: Matching id and describedby looks good! But after testing things out with a screen reader, I think there are a few things missing that would help the itemErrorMessage get read at the right times. (I need to go back and make some of these same refinements to my own PR's form as well.)

If you add `aria-invalid="false"` when your error condition is present, a screen reader will read to a user that the field is invalid, followed by the error message, when the user is focused on that field. Without this attribute, the hint isn't read aloud by my screen reader when focused on the field. 

You could easily update your code to use the right value for this attribute at the right times by doing something like `aria-invalid={!!errorMessage}` or `aria-invalid={Boolean(errorMessage)}`, that way it will automatically be true whenever the error message is not an empty string.

src/components/AddItemForm/AddItemForm.js
```diff
@@ -47,11 +103,13 @@ const AddItemForm = ({ listId }) => {
...
         value={formValues.itemName}
         onChange={handleChange}
         maxLength="100"
         required
       />
+      <span id="itemErrorMessage">{errorMessage}</span>
```
anderswift, line 112: In addition to aria-invalid on the field, it would be a good idea to also add an `aria-live="assertive"` live region somewhere on this page (or role="alert" attribute, but that was buggy for me in JAWS). A live region will immediately read the text within when it changes, so is useful for form error messages because then they are instantly read when the error occurs. Without this, the user needs to click on the text field to understand that there is an error. 

A common pattern is to use a live region to summarize all form errors (may or may not be displayed as visible to sighted users), then also have hints/descriptions for each field as you have now. In this case, since we only have one field that can have an error, I wonder if we can't just add the `aria-live="assertive"` attribute here on this span? It worked for me in testing anyway!

src/components/AddItemForm/AddItemForm.js
```diff
@@ -19,8 +28,55 @@ const AddItemForm = ({ listId }) => {
...
+  // used when comparing entered item and array of db items for duplicates
+  const normalizeInput = (item) => {
+    return item
+      .toLowerCase()
+      .replace(/\W/g, '') // remove non-word characters aka remove punctuation
+      .replace(/ /g, '') // remove spaces
+      .trim(); // remove leading and trailing whitespace (trim must be last)
+  };
+
+  const addItemToDatabase = async () => {
```
anderswift, line 77: I like the choice here to separate this out into another function. 👍

anderswift (approved):
The duplicate item check works great! I left a few comments on accessibility things we can refine for better UX, and some suggestions for where a bit of refactoring might improve readability, but overall I think you did a good job tackling the issue.

src/components/AddItemForm/AddItemForm.js
```diff
@@ -71,8 +69,8 @@ const AddItemForm = ({ listId }) => {
   const normalizeInput = (item) => {
     return item
       .toLowerCase()
-      .replace(/[.,/#!$%^&*;:{}=\-_`~()@[\]|<>+'"/?]/g, '') // remove punctuation
-      .replace(/\s{2,}/g, ' ') // remove extra spaces from removing punctuation
+      .replace(/\W/g, '') // remove non-word characters aka remove punctuation
+      .replace(/ /g, '') // remove spaces
       .trim(); // remove leading and trailing whitespace (trim must be last)
```
nick-zanetti: Per @anderswift 's comment on this, I think the second line here might be unnecessary. When I commented it out and tested it, the spaces were still removed. Are there different cases this line covers we aren't seeing?

src/components/AddItemForm/AddItemForm.js
nick-zanetti: I'm also wondering here if it might be easier to set this state item as an empty string? So you would have: 

`const [errorMessage, setErrorMessage] = useState('');`

Then: 

`const handleChange = (event) => {
    setErrorMessage(''); // if input field changes, reset error message`

nick-zanetti (approved):
Everything worked well for me, great job! I just had a couple of minor questions/suggestions I left as comments. This seems to fulfill all the AC.

src/components/AddItemForm/AddItemForm.js
connietran-dev: We were debating whether to follow what we had did for defaultFormValues, but this is certainly cleaner! 

We've pushed a commit to simply reset the error message to an empty string.

https://github.com/the-collab-lab/tcl-26-smart-shopping-list/pull/28/commits/99db234ffd95649484cb53d100defa57353c92f1

src/components/AddItemForm/AddItemForm.js
```diff
@@ -19,8 +28,55 @@ const AddItemForm = ({ listId }) => {
...
+            formValues.itemName,
+            dbItemArray,
+          );
+
+          // if item exists, show error message, otherwise, continue with adding to database
+          if (duplicateResult === true) {
+            setErrorMessage('Item already exists in Shopping List');
+          } else {
+            addItemToDatabase();
+          }
```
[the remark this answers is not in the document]
connietran-dev: Great suggestion Ander! I like it, and it helps remove an extra function we have to keep track of. We've implemented your suggestion in the commit below. (Next time I will use the GitHub GUI to commit your suggestions. I didn't realize it until after I committed the changes.) Thank you for writing up the code!

https://github.com/the-collab-lab/tcl-26-smart-shopping-list/pull/28/commits/4c0c089217464ae2967835fd0fecc2ffb9a3935a

src/components/AddItemForm/AddItemForm.js
connietran-dev: Good catches. After some late-ish nights, this is why it's helpful to have a fresh set of eyes :) removing `/ / g` and the `trim()` work. See commit here: https://github.com/the-collab-lab/tcl-26-smart-shopping-list/pull/28/commits/4c0c089217464ae2967835fd0fecc2ffb9a3935a

src/components/AddItemForm/AddItemForm.js
yenly: TIL @anderswift Great explanation!

src/components/AddItemForm/AddItemForm.js
yenly: I concur. 👏🏽

yenly (approved):
Fantastic work! These changes met the ACs. I recommend pulling the accessibility changes out to a new issue and create a new PR for those changes. This way it won't be a blocker for this one.

src/components/AddItemForm/AddItemForm.js
Cjstoney: Yes! This is an excellent improvement in readability. Well done 👏🏽

src/components/AddItemForm/AddItemForm.js
Cjstoney: +1 for TIL!

Cjstoney (approved):
This is looks so good! Really impressed with the way you both iterated over the code that you originally wrote to make it more legible, accessible and elegant. Well done!

src/components/AddItemForm/AddItemForm.js
anderswift: Actually, I'm updating my suggestion after a bunch of testing with focus() and different aria attributes. If you take the suggestion to force focus on the field after the form is submitted with an error, the describedby text will be read automatically and function basically like an alert. So because we only have one field that can have an error right now, we don't need a live region for a "big picture" summary, and setting it on the existing describedby message will just make the screen reader read things twice.

abowler2 (approved):
Like the others have said, really nice work!

Issue #31, opened 2021-08-02
Accessibility: Form accessibility updates


connietran-dev:
This issue is to address the accessibility-related feedback on the comments in the PR for AddForm Item: https://github.com/the-collab-lab/tcl-26-smart-shopping-list/pull/28

Also create additional messaging for App.js when creating list.

connietran-dev:
Addressed by PR #33

Issue #32, opened 2021-08-02
Refactor AddItemForm so db is not called on every submit


connietran-dev:
- This should be done after #32 
- Take out database calls into its own function and pass it down. 
- Minor issue that may be solved with this: When adding item, it jumps to top of list, then sorts.

PR #33, opened 2021-08-02
Form errors and accessibility

connietran-dev:
## Description

In AddItemForm.js, we've added improved accessibility for the form. We also added a div with role="alert" to make more general announcements on the form if there are broader connection or errors when adding an item. 

On handleSubmit:
- We check that querySnapshot.empty is NOT empty (line 44). If it is not empty, we proceed with checking for duplicates and adding to the database if it's not.
- If there results ARE empty and data is NOT cached, just add the item (line 58)
- Else, we're assuming a connection issue and so we display an error message (line 61)

Note that on line 70 of AddItemForm.js, the try/catch block is not likely to reach this error, but we have added an error in an abundance of caution :) we also recognize this will probably be refactored lately (issue #32).

In Home.js, we also changed console.log's to error messages that display to the user so they are aware of issues with creating their list.

## Related Issue

Closes Issue https://github.com/the-collab-lab/tcl-26-smart-shopping-list/issues/31 and addresses accessibility-related comments in PR https://github.com/the-collab-lab/tcl-26-smart-shopping-list/pull/28

## Acceptance Criteria

N/A

## Type of Changes

|   | :sparkles: New feature     |
|  ✓  | :hammer: Refactoring       |

### Before

image

### After

(Same user experience but there is better screenreading and accessibility)

image

New `<div role="alert">` for general form errors

image

1. Pull down changes.
2. You can simulate being offline by going to DevTools > Network > Offline (under the No Throttling dropdown). Try creating a new list while offline. A general error will appear in the `<div role="alert">` saying there was a connection error.
3. Go back online. Turn on screenreader. Add an item that already exists. Focus now goes to itemName field since we added a useRef() React Hook on the item field. Screenreader will announce that the item name is invalid since we added `aria-invalid={Boolean(itemErrorMessage)} `.
4.  Simulate being offline. Try adding an item. A general error will appear in the `<div role="alert">` saying there was a connection error.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -32,7 +32,7 @@ function App() {
     return isTokenValid(token).then((listExists) => {
       if (listExists) {
         // if isTokenValid returns true the token is taken, so try again
-        createList();
+        return createList();
       } else {
         // otherwise the token is not already associated with a list, so we can safely use it
         return db
--- a/src/App.js
+++ b/src/App.js
@@ -93,7 +93,7 @@ function App() {
           if (!listExists) localStorage.removeItem('token');
         })
         .catch((error) => {
-          console.log('Error getting list: ', error);
+          console.error('Problem getting list: ', error.message);
         });
     }
   }, []);
--- a/src/components/AddItemForm/AddItemForm.js
+++ b/src/components/AddItemForm/AddItemForm.js
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { useState, useRef } from 'react';
 import { db } from '../../lib/firebase.js';
 
 const AddItemForm = ({ listId }) => {
@@ -10,6 +10,7 @@ const AddItemForm = ({ listId }) => {
 
   /** State **/
   const [formValues, setFormValues] = useState(defaultFormValues);
+  const itemNameRef = useRef(); // ref for the item name field
   const [errorMessage, setErrorMessage] = useState('');
 
   /** Functions **/
@@ -40,9 +41,11 @@ const AddItemForm = ({ listId }) => {
             normalizeInput(doc.data().itemName),
           );
 
-          // if item exists, show error message, otherwise, continue with adding to database
+          // if item exists, show error message and put focus on field
+          // otherwise, continue with adding to database
           if (dbItemArray.includes(normalizeInput(formValues.itemName))) {
-            setErrorMessage('Item already exists in Shopping List');
+            setErrorMessage('Item already exists in Shopping List.');
+            itemNameRef.current.focus();
           } else {
             addItemToDatabase();
           }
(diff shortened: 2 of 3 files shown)


yenly (approved):
Fantastic collaboration! :rocket:

src/components/AddItemForm/AddItemForm.js
```diff
@@ -36,22 +41,35 @@ const AddItemForm = ({ listId }) => {
...
+            setAddItemFormError(
+              'Sorry, there was a problem adding your item. Please check your connection and try again.',
+            );
           }
         });
     } catch (err) {
       console.log(err);
+      setAddItemFormError(
+        'Sorry, there was a problem adding your item. Please check your connection and try again.',
+      );
```
nick-zanetti, line 72: For the errors that are affiliated with the default catch, is it basically safe to assume there is some problem with the user's internet? Is this the error that would get thrown when there is some problem with the database where the `get()` promise rejects? Either way, I think the actual message is probably fine, just want to clarify for my own understanding.

src/App.js
```diff
@@ -32,7 +32,7 @@ function App() {
     return isTokenValid(token).then((listExists) => {
       if (listExists) {
         // if isTokenValid returns true the token is taken, so try again
-        createList();
+        return createList();
```
nick-zanetti, line 35: Curious as to why we change this from a call to a return?

nick-zanetti (approved):
These look good to me! My comments were just questions to help myself understand everything better.

SheilaRMoore (approved):

src/App.js
anderswift: When Connie and I were working on the error handling (therefore trying to trigger those catch zones in the `handleCreateList` function that calls this one), I noticed that the return is omitted in this one area. I'm pretty sure there should always be a return (and that's definitely the way I originally wrote it whoops), because we are expecting the function to return a Promise when we use it in `handleCreateList`. However, I didn't actually reproduce a "problem" caused by the lack of return (since it's not easy to trigger that area unless I specifically add some extra code to ensure duplicate tokens happen often). 

@yenly, @aprilleone or @Cjstoney: maybe one of you could look this over and provide some insight into whether the code could potentially break without that return?

src/components/AddItemForm/AddItemForm.js
anderswift: I honestly can't think of any situation where the catch would run? Since a connection issue doesn't throw any error, and instead is handled in the else statement just above, any other problem would have to be something weird with Firebase which we figured is just as well reported as a vague connection issue. But actually... looking at this again, I think we could neaten it up by throwing an error in the else statement, so we only have to write setting the form error state once.

PR #35, opened 2021-08-03
Check list items as purchased

### LINKED ISSUE #8: 8. As a user, I want to mark an item on my shopping list as purchased so the app can learn how often I buy different items.
AC:

* User is able to tap a checkbox or similar UI element to mark an item in the list as purchased
* Item should be shown as checked for 24 hours after the purchase is made (i.e. we assume the user does not need to buy the item again for at least 1 day)

Notes:

* One challenge your pair may work through is figuring out you they will test this.
* E.g., maybe create a few sample items then go into the Firestore console and manipulate the dates. T
* The wirefrrame shows the use of simple checkboxes for this purpose, which is fine for now. https://github.com/the-collab-lab/smart-shopping-list/blob/main/_resources/wireframes/list-populated.png

anderswift:
## Description

This PR adds functionality to allow a user to check off items on their shopping list. We also prevent user's from checking an item twice within 24 hours. 

### ShoppingListItem component updates
We added a checkbox input and label for items. The component checks if the item was purchased within the last 24 hours using helper function `isUnder24hSincePurchased` to compare timestamps. While the purchase is recent, the checkbox is disabled.

### ShoppingList component updates
We added the `handleCheck` function to this component and passed it to the ShoppingListItem component to run onChange for any checkbox. `handleCheck` makes sure the checkbox is checked, and if so updates the associated item's Firestore document with a current timestamp for the lastPurchaseDate and increments the numberOfPurchases.

## Related Issue

Closes #8

## Acceptance Criteria

- [ ] User is able to tap a checkbox or similar UI element to mark an item in the list as purchased
- [ ] Item should be shown as checked for 24 hours after the purchase is made (i.e. we assume the user does not need to buy the item again for at least 1 day)

## Type of Changes

### Before
image

### After
image

1. Pull down this branch and check it out.
2. Run npm install if you don't have the dependencies installed yet.
3. Run npm start to start local server.
4. Once you have created a list, add some items and check them off.
5. Check Firestore database to verify the timestamp and number of purchases were updated correctly.
6. In Firestore, manipulate the lastPurchaseDate timestamp of an item to verify that you are able to check it off again 24 hours after lastPurchaseDate.

### DIFF
--- a/src/components/AddItemForm/AddItemForm.js
+++ b/src/components/AddItemForm/AddItemForm.js
@@ -108,7 +108,7 @@ const AddItemForm = ({ listId }) => {
         />
         <label
           htmlFor="soonOption"
-          className="add-item-form__label add-item-form__label_type_radio label"
+          className="add-item-form__label add-item-form__label_type_radio label label_check-radio"
         >
           Soon
         </label>
@@ -124,7 +124,7 @@ const AddItemForm = ({ listId }) => {
         />
         <label
           htmlFor="kindaSoonOption"
-          className="add-item-form__label add-item-form__label_type_radio label"
+          className="add-item-form__label add-item-form__label_type_radio label label_check-radio"
         >
           Kind of Soon
         </label>
@@ -140,7 +140,7 @@ const AddItemForm = ({ listId }) => {
         />
         <label
           htmlFor="notSoonOption"
-          className="add-item-form__label add-item-form__label_type_radio label"
+          className="add-item-form__label add-item-form__label_type_radio label label_check-radio"
         >
           Not Soon
         </label>
--- a/src/components/ShoppingList/ShoppingList.js
+++ b/src/components/ShoppingList/ShoppingList.js
@@ -1,3 +1,4 @@
+import firebase from 'firebase/app';
 import { db } from '../../lib/firebase.js';
 import { useCollection } from 'react-firebase-hooks/firestore';
 import ShoppingListItem from '../ShoppingListItem/ShoppingListItem.js';
@@ -6,6 +7,21 @@ function ShoppingList({ listId }) {
   const [listItems, loading, error] = useCollection(
     db.collection(`lists/${listId}/items`).orderBy('purchaseInterval', 'asc'),
   );
+
+  const handleCheck = (e) => {
+    if (!e.target.checked) return; // don't do anything if the checkbox isn't checked
+
+    const itemId = e.target.value;
+    db.collection(`lists/${listId}/items`)
+      .doc(itemId)
+      .update({
+        lastPurchaseDate: firebase.firestore.FieldValue.serverTimestamp(),
+        numberOfPurchases: firebase.firestore.FieldValue.increment(1),
+      })
+      .catch((err) => {
+        console.log(err);
+      });
+  };
 
   return (
     <div className="shopping-list">
@@ -22,9 +38,15 @@ function ShoppingList({ listId }) {
       )}
 
       {!loading && listItems && (
-        <ul className="shopping-list__list">
+        <ul className="shopping-list__list list-reset">
           {listItems.docs.map((doc) => (
-            <ShoppingListItem key={doc.id} item={doc.data()} />
+            <ShoppingListItem
+              key={doc.id}
+              listId={listId}
+              itemId={doc.id}
+              item={doc.data()}
+              handleCheck={handleCheck}
+            />
           ))}
         </ul>
       )}
--- a/src/components/ShoppingList/ShoppingList.js
+++ b/src/components/ShoppingList/ShoppingList.js
@@ -9,8 +9,6 @@ function ShoppingList({ listId }) {
   );
 
   const handleCheck = (e) => {
-    if (!e.target.checked) return; // don't do anything if the checkbox isn't checked
-
     const itemId = e.target.value;
     db.collection(`lists/${listId}/items`)
       .doc(itemId)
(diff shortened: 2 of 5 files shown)


src/components/ShoppingListItem/ShoppingListItem.js
```diff
@@ -1,5 +1,42 @@
...
+  // update whether item is recently purchased
+  useEffect(() => {
+    // make sure properties exist and are not null
+    if (
+      'lastPurchaseDate' in item &&
+      item.lastPurchaseDate &&
+      'seconds' in item.lastPurchaseDate
+    )
+      setRecentlyPurchased(
+        isUnder24hSincePurchased(item.lastPurchaseDate.seconds),
```
connietran-dev, line 17: Is there a reason for using .seconds here?

src/components/AddItemForm/AddItemForm.js
```diff
@@ -140,7 +140,7 @@ const AddItemForm = ({ listId }) => {
         />
         <label
           htmlFor="notSoonOption"
-          className="add-item-form__label add-item-form__label_type_radio label"
+          className="add-item-form__label add-item-form__label_type_radio label label_check-radio"
```
connietran-dev, line 143: Is the additional class just for the future?

src/components/ShoppingList/ShoppingList.js
```diff
@@ -7,6 +8,21 @@ function ShoppingList({ listId }) {
...
+  const handleCheck = (e) => {
+    if (!e.target.checked) return; // don't do anything if the checkbox isn't checked
+
+    const itemId = e.target.value;
+    db.collection(`lists/${listId}/items`)
+      .doc(itemId)
+      .update({
+        lastPurchaseDate: firebase.firestore.FieldValue.serverTimestamp(),
+        numberOfPurchases: firebase.firestore.FieldValue.increment(1),
+      })
```
connietran-dev, line 20: I like the .update here!

src/components/ShoppingListItem/ShoppingListItem.js
```diff
@@ -1,5 +1,42 @@
...
+  }, [item]);
+
+  return (
+    <li className="shopping-list__item item">
+      <input
+        id={`item-${itemId}`}
+        value={itemId}
+        type="checkbox"
+        disabled={recentlyPurchased}
+        checked={recentlyPurchased}
```
connietran-dev, line 28: This is really nice how you have set this using state and then setting these attributes!

connietran-dev (approved):
Great job. The functionality works great. I was able to edit the timestamp in Firestore, and the checkboxes became enabled again.

I had just one question about why handleCheck was declared in ShoppingList.js and then passed down to ShoppingListItem.js as a prop. But this is something I'd like to bring up in office hours with @yenly or any of the other mentors. It is just something I'd like to understand about React and best practice there. 

Otherwise, great job!

nick-zanetti (approved):
This looks good to me! I do have some thoughts on this that I am moving to a discussion since I don't believe they are critical to address with this PR.

src/components/ShoppingListItem/ShoppingListItem.js
anderswift: What actually gets returned is a Firebase object (try out logging it to the console and you'll see what I mean). To get the actual seconds (as in Unix epoch seconds if I'm understanding correctly), we needed to use the seconds property.

src/components/AddItemForm/AddItemForm.js
anderswift: Yes, we added this because we thought we'd want a generic class for labels for checkbox/radio buttons, since we will almost certainly style them similarly (as in beside the check/radio, instead of on a separate line).

src/components/ShoppingList/ShoppingList.js
yenly: +1 on the `.update` Excellent choice in using firebase functions to update these values.

src/components/ShoppingListItem/ShoppingListItem.js
```diff
@@ -1,5 +1,42 @@
...
+  const [recentlyPurchased, setRecentlyPurchased] = useState(false);
+
+  // update whether item is recently purchased
+  useEffect(() => {
+    // make sure properties exist and are not null
+    if (
+      'lastPurchaseDate' in item &&
+      item.lastPurchaseDate &&
+      'seconds' in item.lastPurchaseDate
+    )
```
yenly, line 15: Not a must change. [Optional chaining is fully supported now](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining). 
```suggestion
    if (item?.lastPurchaseDate?.seconds)
```

yenly (approved):
Awesome work! Excellent choices in using firebase functions for timestamps/increments as well as creating a helper function to calculate time.

src/components/ShoppingList/ShoppingList.js
```diff
@@ -7,6 +8,21 @@ function ShoppingList({ listId }) {
     db.collection(`lists/${listId}/items`).orderBy('purchaseInterval', 'asc'),
   );
 
+  const handleCheck = (e) => {
+    if (!e.target.checked) return; // don't do anything if the checkbox isn't checked
```
abowler2, line 12: I don't think this is needed since `handleCheck` will only fire when the checkbox is checked.

abowler2 (approved):
Nice job! 

Compliments again on understanding your tools and utilizing the methods that they give you with the use of `.update` from Firebase.

src/components/ShoppingListItem/ShoppingListItem.js
anderswift: 🤯

Cjstoney (approved):

PR #39, opened 2021-08-10
Update purchase interval when checking off item

### LINKED ISSUE #10: 10. As an item, I want my estimated next purchase date to be computed at the time my purchase is recorded in the database so the app can learn how often I buy different items.
AC:

* When a purchase is recorded, the estimated number of days until the next purchase date should be calculated and recorded in the database

The script at `/src/lib/estimates.js` should be used to calculate the next estimated purchase interval.

Notes:

* Think through the data that is needed as inputs to the function in `estimates.js` and how to use the output.
* It is likely at this point you'll need to adjust the schema for an item to accomodate the data points required.

anderswift:
## Description

This PR adds functionality to save an updated estimated purchase interval whenever an item is purchased. We do this by calling the `calculateEstimate` function (from `/lib/estimates.js`) to calculate the purchase interval estimate. This action is performed and the result saved within the `ShoppingList `component's `checkAsPurchased `function, which fires when the user checks off an item as purchased.

A few notes:
- `calculateEstimate` requires 3 parameters: `lastEstimate` (last estimated purchase interval), `latestInterval` (days between the most recent and previous purchase) and `numberOfPurchases`

- We can provide the `lastEstimate` and `numberOfPurchases` from the item's current data, but need to calculate `latestInterval` from the item's `lastPurchaseDate` and the current time. We added the Luxon date library and converted dates to Luxon's DateTime object to handle this.

- As instructed by a prior issue's AC, when an item is first saved to the Firestore database, the `lastPurchaseDate` is set to null because there's no purchase history yet. Since it's impossible to calculate a `latestInterval` between the current date and a null value, we ensure there will always be a default value used as the `latestInterval` in this case. We use the item's current purchaseInterval as the default value, which is initially set by the user when they add the item (7, 14, or 30 days).

## Related Issue

Closes #10 : As an item, I want my estimated next purchase date to be computed at the time my purchase is recorded in the database so the app can learn how often I buy different items.

## Acceptance Criteria

- [ ] When a purchase is recorded, the estimated number of days until the next purchase date should be calculated and recorded in the database

## Type of Changes

| ✓ | :link: Update dependencies |

1. Pull branch and npm install to add the Luxon package.
2. Take note of an item's purchaseInterval in the database.
3. Check off that item in the UI.
4. The `purchaseInterval` for that item should update in the database, according to the formula in the `calculateEstimate` function. 
(Note: the first time an item is checked off as purchased, the `purchaseInterval` should not change because there's not enough purchase history to calculate a meaningful estimate. When an item has been purchased at least once before, the `purchaseInterval` should usually change. You may need to manipulate the `lastPurchaseDate` or `numberOfPurchases` in Firestore in order to observe how the `purchaseInterval` updates in different scenarios.)

### DIFF
--- a/src/components/ShoppingList/ShoppingList.js
+++ b/src/components/ShoppingList/ShoppingList.js
@@ -3,6 +3,8 @@ import { db } from '../../lib/firebase.js';
 import { useCollection } from 'react-firebase-hooks/firestore';
 import ShoppingListItem from '../ShoppingListItem/ShoppingListItem.js';
 import { useHistory } from 'react-router-dom';
+import calculateEstimate from '../../lib/estimates.js';
+import { DateTime } from 'luxon';
 
 function ShoppingList({ listId }) {
   let history = useHistory();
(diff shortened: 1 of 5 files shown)


src/components/ShoppingList/ShoppingList.js
```diff
@@ -11,13 +13,41 @@ function ShoppingList({ listId }) {
...
+      lastPurchaseDate === null
+        ? item.purchaseInterval
+        : getLatestInterval({ lastPurchaseDate, newPurchaseDate });
+
+    const newPurchaseInterval = calculateEstimate(
+      item.purchaseInterval,
+      latestInterval,
+      item.numberOfPurchases,
+    );
+
```
nick-zanetti, line 41: This was a really helpful update, especially with the updated comments, great job!

src/components/ShoppingList/ShoppingList.js
```diff
@@ -11,13 +13,41 @@ function ShoppingList({ listId }) {
     db.collection(`lists/${listId}/items`).orderBy('purchaseInterval', 'asc'),
   );
 
-  const handleCheck = (e) => {
-    const itemId = e.target.value;
+  // Helper function to get the latest interval between purchases (expects Luxon date objects)
+  const getLatestInterval = ({ lastPurchaseDate, newPurchaseDate }) => {
```
nick-zanetti, line 17: So for my own understanding, why do we need to pass our args into this function with {}? I'm used to doing this when we are destructuring out something from an object directly to save using dot notation in the function body, but that doesn't seem to be what we are doing here. Do we need to do this because both variables represent Luxon date objects?

nick-zanetti (changes requested):
Currently, this doesn't seem to be behaving as expected for me. When I check off an item, the "lastPurchaseDate" updates in the database, but not the purchaseInterval

UPDATE: Actually, this is working. I see that for `calculateEstimate()` to start working, `numberOfPurchases` must be >=2. So in this build you need to check the item off, change the date to greater than 24 hours so it becomes unchecked, then re-check it again so that `numberOfPurchases` becomes 2 and `calculateEstimate()` runs that part of the conditional. You should probably update the testing instructions to reflect this.

src/components/ShoppingList/ShoppingList.js
anderswift: There's no reason it needs to be done this way, and is unrelated to Luxon. This is just a style of handling parameters that I've been encouraged to use when the order of the parameters might not be intuitive, or there are just a lot of them so confusion is easy. (It's also a really common pattern in some of the codebases I work with all the time, which use loosely typed languages like JS and PHP.) But we could just as well write it like this: 
`const getLatestInterval = (lastPurchaseDate, newPurchaseDate) => { ... }`

The one advantage of using a single object parameter and destructuring is that you can then call the function with the "arguments" (properties of one argument really) in any order, and because you've explicitly named them it will still work the same way. For example, as is either of these function calls would return the same output:
```
getLatestInterval({ newPurchaseDate, lastPurchaseDate });
getLatestInterval({ lastPurchaseDate, newPurchaseDate });
```

nick-zanetti (approved):

abowler2 (approved):
Nice job digging into the `calculateEstimate` function and working out how to handle the null value on a new item. And thanks for bringing your questions about how to optimize your solution to office hours, it was a lot of fun refactoring with you all.

src/components/ShoppingList/ShoppingList.js
yenly: 👏🏽

yenly (approved):
Fantastic work. The comments and how you structure the code for calculating the numbers needed to get the estimated is very well organized.

Cjstoney (approved):
Really well done. I love how organized this pull request is for a pretty complex problem. The comments throughout are very helpful as well.  👏🏽

PR #41, opened 2021-08-11
Add filter input to the shopping list

### LINKED ISSUE #9: 9. As a user, I want to filter my shopping list to make it easier to locate an item in the list.
AC:

* Display a text field above the top of the shopping list
* As the user types into the field, the list should narrow to display only items that contain the text the user entered in the filter field
* When the field has text in it, the user should be able to tap a UI element (e.g., with an "X" button next to the field) to clear the field
* The filter text should match any part of the item name (i.e. it should _not_ only match from the start of the string)

Notes:

* The wireframe shows the filter field, but not it being used. https://github.com/the-collab-lab/smart-shopping-list/blob/main/_resources/wireframes/list-populated.png
* The following example app from MDN is a great example of providing feedback that is accessible to screenreaders when things change on the screen. https://moz-todomvc-react.netlify.app/

nick-zanetti:
## Description

This is a fairly simple commit which adds a field to allow the user to filter their shopping list through a text field at the top. 

### ShoppingList.js

I elected to to simply call a `.filter()` in-line where we render our `listItems`. This filters the items based on the new `filter` state. The filter input field is made as a controlled input, with `value` being set to the `filter` state and a `handleInput` function which sets `filter` to the user's input. 

There is also a "clear filter" button which will clear out the filter input field. 

## Related Issue

Closes #9 

## Acceptance Criteria

- [ ] Display a text field above the top of the shopping list
- [ ] As the user types into the field, the list should narrow to display only items that contain the text the user entered in the filter field
- [ ] When the field has text in it, the user should be able to tap a UI element (e.g., with an "X" button next to the field) to clear the field
- [ ] The filter text should match any part of the item name (i.e. it should not only match from the start of the string)

## Type of Changes

1. Pull down the branch
2. Run `npm start` to start the dev. server
3. Add several items to a shopping list
4. Type in the filter box, and make sure the list immediately responds, and that items show up if they match any part of the input

### DIFF
--- a/src/components/ShoppingList/ShoppingList.js
+++ b/src/components/ShoppingList/ShoppingList.js
@@ -47,49 +47,50 @@ function ShoppingList({ listId }) {
       );
     } else {
       return (
-        <ul className="shopping-list__list list-reset">
-          {listItems.docs
-            .filter((doc) => {
-              return doc.data().itemName.includes(filter.toLowerCase());
-            })
-            .map((doc) => (
-              <ShoppingListItem
-                key={doc.id}
-                listId={listId}
-                itemId={doc.id}
-                item={doc.data()}
-                handleCheck={handleCheck}
-              />
-            ))}
-        </ul>
+        <div>
+          <div className="filter">
+            <label htmlFor="filterInput" className="filter__label label">
+              Filter items
+            </label>
+            <input
+              type="text"
+              id="filterInput"
+              name="filterInput"
+              value={filter}
+              onChange={handleInput}
+              className="filter__text-field text-field"
+            />
+            <button
+              type="button"
+              aria-label="clear"
+              className="filter__button"
+              onClick={() => setFilter('')}
+            >
+              Clear Filter
+            </button>
+          </div>
+          <ul className="shopping-list__list list-reset">
+            {listItems.docs
+              .filter((doc) =>
+                new RegExp(filter, 'i').test(doc.data().itemName),
+              )
+              .map((doc) => (
+                <ShoppingListItem
+                  key={doc.id}
+                  listId={listId}
+                  itemId={doc.id}
+                  item={doc.data()}
+                  handleCheck={handleCheck}
+                />
+              ))}
+          </ul>
+        </div>
       );
     }
   };
 
   return (
     <div className="shopping-list">
-      <div className="filter">
-        <label htmlFor="filterInput" className="filter__label">
-          Filter items
-        </label>
-        <input
-          type="text"
-          id="filterInput"
-          name="filterInput"
-          value={filter}
-          onChange={handleInput}
-          className="filter__text-field text-field"
-        />
-        <button
-          type="button"
-          aria-label="clear"
-          className="filter__button"
-          onClick={() => setFilter('')}
-        >
-          Clear Filter
-        </button>
-      </div>
-
       {loading && (
         <div className="shopping-list__notice notice notice_type_loading">
           Loading...
--- a/src/components/ShoppingList/ShoppingList.js
+++ b/src/components/ShoppingList/ShoppingList.js
@@ -3,6 +3,7 @@ import { db } from '../../lib/firebase.js';
 import { useCollection } from 'react-firebase-hooks/firestore';
 import ShoppingListItem from '../ShoppingListItem/ShoppingListItem.js';
 import { useHistory } from 'react-router-dom';
+import { useState } from 'react';
 
 function ShoppingList({ listId }) {
   let history = useHistory();
@@ -10,6 +11,8 @@ function ShoppingList({ listId }) {
   const [listItems, loading, error] = useCollection(
     db.collection(`lists/${listId}/items`).orderBy('purchaseInterval', 'asc'),
   );
+
+  const [filter, setFilter] = useState('');
 
   const handleCheck = (e) => {
     const itemId = e.target.value;
@@ -22,6 +25,10 @@ function ShoppingList({ listId }) {
       .catch((err) => {
         console.log(err);
       });
+  };
+
+  const handleInput = (e) => {
+    setFilter(e.target.value);
   };
 
   const handleClick = () => {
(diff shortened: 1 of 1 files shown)


src/components/ShoppingList/ShoppingList.js
```diff
@@ -41,22 +48,48 @@ function ShoppingList({ listId }) {
...
+        />
+        <button
+          type="button"
+          aria-label="clear"
+          className="filter__button"
+          onClick={() => setFilter('')}
+        >
+          Clear Filter
+        </button>
+      </div>
```
anderswift, line 91: Since this isn't needed when there are no list items yet, what do you think of moving this to the `createListElement` function, just above the item list? That way the UI will be more clean and straightforward for a new user.

src/components/ShoppingList/ShoppingList.js
```diff
@@ -41,22 +48,48 @@ function ShoppingList({ listId }) {
...
+            ))}
         </ul>
       );
     }
   };
 
   return (
     <div className="shopping-list">
+      <div className="filter">
+        <label htmlFor="filterInput" className="filter__label">
```
anderswift, line 72: I missed this when you asked me earlier, but to have it pick up our generic label styles we should put an additional class of simply `label` here too.

src/components/ShoppingList/ShoppingList.js
```diff
@@ -41,22 +48,48 @@ function ShoppingList({ listId }) {
...
-              listId={listId}
-              itemId={doc.id}
-              item={doc.data()}
-              handleCheck={handleCheck}
-            />
-          ))}
+          {listItems.docs
+            .filter((doc) => {
+              return doc.data().itemName.includes(filter.toLowerCase());
+            })
```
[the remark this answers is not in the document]
nick-zanetti: This works great, thanks! Got a quick lesson on the RegExp constructor as well.

src/components/ShoppingList/ShoppingList.js
nick-zanetti: Sure, I just tried it and I'm definitely in favor of having as little information as necessary on the screen at once for a new user

src/components/ShoppingList/ShoppingList.js
```diff
@@ -24,6 +27,10 @@ function ShoppingList({ listId }) {
       });
   };
 
+  const handleInput = (e) => {
```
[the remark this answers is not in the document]
nick-zanetti: Since our updater function for `filter` is called `setFilter`, I'd actually prefer to keep this one how it is.

src/components/ShoppingList/ShoppingList.js
```diff
@@ -40,17 +47,44 @@ function ShoppingList({ listId }) {
...
+              name="filterInput"
+              value={filter}
+              onChange={handleInput}
+              className="filter__text-field text-field"
             />
-          ))}
-        </ul>
+            <button
+              type="button"
+              aria-label="clear"
```
connietran-dev, line 65: Do we want the`aria-label` to also read "clear filter" since the text within the button is also "Clear Filter"? When I turn on a screenreader, it reads "clear, button" whereas if we add "clear filter" for the `aria-label` it reads "clear filter button". I think it'd be more clear.

connietran-dev (approved):
Nice changes based on our feedback. Overall, functionality works, and the code is clean and easy to read. Great job

src/components/ShoppingList/ShoppingList.js
anderswift: Yes, I agree, or even "Clear item filter". 

Also want to note, because I may not have communicated this clearly earlier: We don't currently need an aria-label *and* text inside the button, one or the other is fine. Since we discussed down the line removing the text inside the button tags and making it an image instead, and in that case *would* need the aria-label, it doesn't hurt to keep it. But I just want to make sure we're clear on when an aria-label is necessary and when it's not—screen readers will read inner text of a button, and an aria-label will override that text if it's provided.

nick-zanetti: That makes sense, and yes I think I just left the tex in there for now for the purposes of getting through this review cycle, and then we can remove it later once we add that styling.

abowler2 (approved):
Nicely done!

src/components/ShoppingList/ShoppingList.js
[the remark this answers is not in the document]
Cjstoney: @nick-zanetti is on the money with this one. I learned during a technical interview that with useState, best practice is to have the setter function be named 'set<what ever is being set>'. Always a good instinct to want specific naming though.

Cjstoney (approved):
This looks great!

src/components/ShoppingList/ShoppingList.js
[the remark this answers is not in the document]
anderswift: @nick-zanetti & @Cjstoney, to clarify I definitely agree we should use filter and setFilter and those are perfect naming choices. 👌 My only thought was that, especially once our PRs are merged, we have multiple functions in this component that "handle input" from form fields (and could potentially end up with more down the line). So just thought "handleInput" as a function name could be a little ambiguous, but I also take your point Nick about too many things being called filter. 

Given that it's a tiny function, we could get around the naming entirely and make its use 100% clear by making it an anonymous function like you did with the clear button onClick. But I'm not gonna fight you on it. 😄

src/components/ShoppingList/ShoppingList.js
```diff
@@ -40,17 +47,44 @@ function ShoppingList({ listId }) {
...
+              onClick={() => setFilter('')}
+            >
+              Clear Filter
+            </button>
+          </div>
+          <ul className="shopping-list__list list-reset">
+            {listItems.docs
+              .filter((doc) =>
+                new RegExp(filter, 'i').test(doc.data().itemName),
+              )
```
yenly, line 76: I typically see the use of array `.include()` method but this is also very nice!

yenly (approved):
Looks like you knock it out the park!

src/components/ShoppingList/ShoppingList.js
anderswift: @yenly The regex came about as result of a mob session idea to make the filter case-insensitive.

PR #42, opened 2021-08-11
Document titles

### LINKED ISSUE #36: As a user using a screen reader, I want to receive feedback when navigating to a different page so that I always understand what context I'm in.
This could be handled by changing the HTML document title for different views.

anderswift:
## Description

This PR adds React Helmet, and uses it to set unique document titles for each page. 

## Related Issue

Closes #36 

## Type of Changes

| ✓ | :link: Update dependencies |

1. Run npm install.
2. Navigate to different pages and review document titles.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13230,10 +13230,26 @@
       "resolved": "https://registry.npmjs.org/react-error-overlay/-/react-error-overlay-6.0.9.tgz",
       "integrity": "sha512-nQTTcUu+ATDbrSD1BZHr5kgSD4oF8OFjxun8uAaL8RwPBacGBNPf/yAuVVdx17N8XNzRDMrZ9XcKZHCjPW+9ew=="
     },
+    "react-fast-compare": {
+      "version": "3.2.0",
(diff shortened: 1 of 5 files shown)


nick-zanetti (approved):
Looks good! So for our purposes, does this just ensure that the title of the page will be constant whether the user is navigating through a nested component or not?

anderswift:
> Looks good! So for our purposes, does this just ensure that the title of the page will be constant whether the user is navigating through a nested component or not?

Adding Helmet lets us change the [HTML document title](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/title) on different "pages," like the View List or Add Item pages (since our React App is set up SPA-style, by default this title would never change). Having different titles on different pages is helpful for screen reader users, because these titles are read aloud when a new page loads, helping the user to understand where they ended up after clicking navigation elements. (Previously, clicking any navigation element provided no feedback at all to a screen reader user.) 

I believe the way Helmet works is, at any given time, it constructs the document head using whatever's in the Helmet tags of the deepest nested component that's currently rendered. That gives you the capability to override the document head from within any component—very handy!

yenly (approved):
👏🏽

connietran-dev (approved):

PR #43, opened 2021-08-17
Sort list items

### LINKED ISSUE #12: 12. As a user, I want to view a list of my shopping list items in order of how soon I am likely to need to buy each of them again so that it’s clear what I need to buy soon.
Possible item states are as follows:

* Need to buy soon (fewer than 7 days)
* Need to buy kind of soon (between 7 & 30 days)
* Need to buy not soon (more than 30 days)
* Inactive (when there’s only 1 purchase in the database or the purchase is really out of date [the time that has elapsed since the last purchase is 2x what was estimated])

AC:

* Items in the list are shown as visually distinct (e.g., with a different background color on the list item) according to how soon the item is expected to be bought again: Soon, Kind of soon, Not soon, Inactive
* Items should be sorted by the estimated number of days until next purchase
* Items with the same number of estimated days until next purchase should be sorted alphabetically
* Items in the different states should be described distinctly when read by a screen reader

Notes:

* The criteria for an item being inactive has tripped up teams in the past. It's been re-worded to be more clear, but the gist is that an item is inactive if the time that has elapsed since its last purchase is twice (or more) what was estimated the last time it was purchased.
* E.g., if the last estimate was 14 days and 28 days have elapsed, the item is considered inactive.
* Teams in the past have also overengineered the solution here by breaking the list out into separate groups, but it should be achievable in a single list where the visual treatment is different based on what category the item falls into.
* Be sure to not only to use color to denote how soon the user will need to buy the item, as noted in the wireframe. https://github.com/the-collab-lab/smart-shopping-list/blob/main/_resources/wireframes/list-populated.png
* Good luck, you got this!

anderswift:
## Description
This PR adds functionality to sort Shopping List items in order of how soon they are expected to be purchased, with items that seem to be inactive sorted to the bottom. 

Note: We decided that it seemed counterintuitive to have new items immediately be considered 'inactive' and sorted to the bottom of the list, but since new items have never been purchased, there is no `lastPurchaseDate` to use to calculate when the next purchase is expected. To handle this, we made a minor change to the data structure adding a `createdAt` timestamp, and for new items we calculate their days until next purchase and inactivity based on that date instead.

### `getDaysToPurchase`

The `getDaysToPurchase()` function estimates how many days remain until the application expects the user to purchase an item again. If the item has been purchased before, next purchase date is `[purchaseInterval]` days from the last purchase date. If there is no purchase history, the function estimates that the item will be bought `[purchaseInterval]` days from when it was first added to the list.

### `sortListItems` 

The `sortListItems()` function is a [compare function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) which we pass into a `.sort()` call on our array of items. The function sorts items based first on their "status", a new property for each item which is based on the days to purchase (calculated by `getDaysToPurchase()`). After the days to purchase, the function sorts the items alphabetically using [.localeCompare.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) 

### `isItemInactive` 

The `isItemInactive()` function determines whether an item is "inactive" based on the parameters laid out in the AC (there’s only 1 purchase in the database or the purchase is really out of date [the time that has elapsed since the last purchase is 2x what was estimated]). 

## ShoppingListItem.js

We added a `<span>`, which is visually hidden, for content that will be read by screen readers. The text that is read is based on the item status. Finally, set the class of the `<label>` to be conditionally based on `item.status`, and styled the relevant classes with different colors to make the items visually distinct, based on their "status". 

## Related Issue

Closes #12. 

## Acceptance Criteria

- [x] Items in the list are shown as visually distinct (e.g., with a different background color on the list item) according to how soon the item is expected to be bought again: Soon, Kind of soon, Not soon, Inactive
- [x] Items should be sorted by the estimated number of days until next purchase
- [x] Items with the same number of estimated days until next purchase should be sorted alphabetically
- [x] Items in the different states should be described distinctly when read by a screen reader

## Type of Changes

### Before

image

### After

image

1. Pull down the branch, and run npm install if you haven't yet installed the dependencies.
3. Make sure you are using a list with a variety of items with different purchase histories and calculated purchase intervals to best observe sorting. You can join the list with token 'watts burly enrico' to meet this requirement, and/or manipulate data in the firestore console.
4. Compare the items in the firestore database, and check whether they are sorting according to the following rules:
   - Items are sorted by the estimated number of days until next purchase (the difference between the current date and the `lastPurchaseDate`, or `createdAt` for new items, plus `[purchaseInterval]` days, rounded to the nearest day)
   - Items expected to be purchased sooner appear higher up
   - If more than 2 x `[purchaseInterval]` days have passed since an item's `lastPurchaseDate` (or `createdAt` date for new items), it is inactive and should be sorted to the bottom of the list
   - Items with the same estimated number of days until purchase are sorted alphabetically
5. Make sure the items have an appropriately coloured label and hidden screen reader text indicating these different possible item states: 
   - Need to buy soon (fewer than 7 days)
   - Need to buy kind of soon (between 7 & 30 days)
   - Need to buy not soon (more than 30 days)
   - Inactive (the time that has elapsed since the last purchase is 2x what was estimated)

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -87,3 +87,13 @@ BEM conventions we are using:
   margin: 0;
   list-style: none;
 }
+
+.visually-hidden {
+  clip: rect(0 0 0 0);
+  clip-path: inset(50%);
+  height: 1px;
+  overflow: hidden;
+  position: absolute;
+  white-space: nowrap;
+  width: 1px;
+}
--- a/src/components/AddItemForm/AddItemForm.js
+++ b/src/components/AddItemForm/AddItemForm.js
@@ -1,4 +1,6 @@
 import { useState, useRef } from 'react';
+
+import firebase from 'firebase/app';
 import { db } from '../../lib/firebase.js';
 
 const AddItemForm = ({ listId }) => {
@@ -82,6 +84,7 @@ const AddItemForm = ({ listId }) => {
       purchaseInterval: Number(formValues.purchaseInterval),
       lastPurchaseDate: null,
       numberOfPurchases: 0,
+      createdAt: firebase.firestore.FieldValue.serverTimestamp(),
     };
 
     try {
--- a/src/components/ShoppingList/ShoppingList.js
+++ b/src/components/ShoppingList/ShoppingList.js
@@ -86,6 +86,7 @@ function ShoppingList({ listId }) {
     const currentDate = DateTime.fromSeconds(Math.floor(Date.now() / 1000));
 
     if (item.lastPurchaseDate?.seconds) {
+      // if an item has been purchased before, compare the time elapsed since lastPurchaseDate to the purchaseInterval
       return (
         currentDate
           .diff(DateTime.fromSeconds(item.lastPurchaseDate.seconds), ['days'])
@@ -93,6 +94,7 @@ function ShoppingList({ listId }) {
         2 * item.purchaseInterval
       );
     } else if (item.createdAt?.seconds) {
+      // if an item has never been purchased, do the same calculating using createdAt date instead of lastPurchaseDate
       return (
         currentDate
           .diff(DateTime.fromSeconds(item.createdAt.seconds), ['days'])
@@ -131,6 +133,8 @@ function ShoppingList({ listId }) {
     if (item.daysToPurchase > 30) return 'not-soon';
   };
 
+  // format the data from firestore into a sorted, filtered array of items
+  // add new properties: item.status and item.daysToPurchase
   const itemsToDisplay = listItems?.docs
     .filter((doc) => new RegExp(filter, 'i').test(doc.data().itemName))
     .map((doc) => {
--- a/src/components/ShoppingList/ShoppingList.js
+++ b/src/components/ShoppingList/ShoppingList.js
@@ -82,14 +82,41 @@ function ShoppingList({ listId }) {
     return daysRemaining.as('days');
   };
 
+  const isItemInactive = (item) => {
+    const currentDate = DateTime.fromSeconds(Math.floor(Date.now() / 1000));
+
+    if (item.lastPurchaseDate?.seconds) {
+      return (
+        currentDate
+          .diff(DateTime.fromSeconds(item.lastPurchaseDate.seconds), ['days'])
+          .as('days') >=
+        2 * item.purchaseInterval
+      );
+    } else if (item.createdAt?.seconds) {
+      return (
+        currentDate
+          .diff(DateTime.fromSeconds(item.createdAt.seconds), ['days'])
+          .as('days') >=
+        2 * item.purchaseInterval
+      );
+    } else return null;
+  };
+
   const sortListItems = (docOne, docTwo) => {
     const itemOne = docOne.data();
     const itemTwo = docTwo.data();
     const daysToPurchaseItemOne = getDaysToPurchase(itemOne);
     const daysToPurchaseItemTwo = getDaysToPurchase(itemTwo);
+    const itemOneInactive = isItemInactive(itemOne);
+    const itemTwoInactive = isItemInactive(itemTwo);
 
-    if (daysToPurchaseItemOne < daysToPurchaseItemTwo) return -1;
-    if (daysToPurchaseItemOne > daysToPurchaseItemTwo) return 1;
+    if (itemOneInactive === itemTwoInactive) {
+      if (daysToPurchaseItemOne < daysToPurchaseItemTwo) return -1;
+      if (daysToPurchaseItemOne > daysToPurchaseItemTwo) return 1;
+    } else {
+      if (itemTwoInactive) return -1;
+      if (itemOneInactive) return 1;
+    }
   };
 
   const createListElement = () => {
@@ -140,7 +167,6 @@ function ShoppingList({ listId }) {
                   itemId={doc.id}
                   item={doc.data()}
                   checkAsPurchased={checkAsPurchased}
-                  daysToPurchase={getDaysToPurchase(doc.data())}
                 />
               ))}
           </ul>
--- a/src/components/ShoppingList/ShoppingList.js
+++ b/src/components/ShoppingList/ShoppingList.js
@@ -17,6 +17,83 @@ function ShoppingList({ listId }) {
 
   const [filter, setFilter] = useState('');
 
+  const currentDate = DateTime.fromSeconds(Math.floor(Date.now() / 1000));
+
+  // format the data from firestore into a sorted, filtered array of items
+  // add new properties: item.status and item.daysToPurchase
+  const itemsToDisplay = listItems?.docs
+    .filter((doc) => new RegExp(filter, 'i').test(doc.data().itemName))
+    .map((doc) => {
+      const item = doc.data();
+      item.id = doc.id;
+      item.daysToPurchase = getDaysToPurchase(item);
+      item.status = getItemStatus(item);
+      return item;
+    })
+    .sort(sortListItems);
+
+  function getDaysToPurchase(item) {
+    let nextPurchaseDate;
+    if (item.lastPurchaseDate?.seconds) {
+      // if the item has been purchased before, next purchase date is `purchaseInterval` days from the lastPurchaseDate
+      nextPurchaseDate = DateTime.fromSeconds(
+        item.lastPurchaseDate.seconds,
+      ).plus({ days: item.purchaseInterval });
+    } else if (item.createdAt?.seconds) {
+      // if there's no purchase history, estimate it will be bought `purchaseInterval` days from when item was created
+      // (user provides this info at item creation)
+      nextPurchaseDate = DateTime.fromSeconds(item.createdAt.seconds).plus({
+        days: item.purchaseInterval,
+      });
+    } else return null;
+
+    const daysRemaining = nextPurchaseDate.diff(currentDate, ['days']);
+    return Math.round(daysRemaining.as('days'));
+  }
+
+  function getItemStatus(item) {
+    if (isItemInactive(item)) return 'inactive';
+    if (item.daysToPurchase < 7) return 'soon';
+    if (item.daysToPurchase >= 7 && item.daysToPurchase <= 30)
+      return 'kind-of-soon';
+    if (item.daysToPurchase > 30) return 'not-soon';
+  }
+
+  function isItemInactive(item) {
+    const estimateDate =
+      item.lastPurchaseDate?.seconds || item.createdAt?.seconds;
+
+    if (estimateDate) {
+      return (
+        currentDate
+          .diff(DateTime.fromSeconds(estimateDate), ['days'])
+          .as('days') >=
+        2 * item.purchaseInterval
+      );
+    }
+  }
+
+  function sortListItems(itemA, itemB) {
+    // if items being compared are both inactive, or neither is inactive, sort by days until next purchase
+    if (
+      (itemA.status === 'inactive' && itemB.status === 'inactive') ||
+      (itemA.status !== 'inactive' && itemB.status !== 'inactive')
+    ) {
+      if (itemA.daysToPurchase < itemB.daysToPurchase) return -1;
+      if (itemA.daysToPurchase > itemB.daysToPurchase) return 1;
+
+      // if we've made it this far, days to purchase is the same for both items so alphabetize
+      return itemA.itemName.localeCompare(itemB.itemName, 'en', {
+        sensitivity: 'base',
+        ignorePunctuation: true,
+      });
+    } else {
+      // if one item is inactive and the other is not, bump down the inactive item
+      if (itemA.status === 'inactive') return 1;
+      if (itemB.status === 'inactive') return -1;
+    }
+  }
+
   // Helper function to get the latest interval between purchases (expects Luxon date objects)
   const getLatestInterval = ({ lastPurchaseDate, newPurchaseDate }) => {
     const duration = newPurchaseDate.diff(lastPurchaseDate, ['days']);
@@ -28,7 +105,7 @@ function ShoppingList({ listId }) {
     const lastPurchaseDate = item.lastPurchaseDate?.seconds
       ? DateTime.fromSeconds(item.lastPurchaseDate.seconds)
       : null; // for new items, lastPurchaseDate will be null so keep it null
-    const newPurchaseDate = DateTime.fromSeconds(Math.floor(Date.now() / 1000));
+    const newPurchaseDate = currentDate;
 
     // if lastPurchaseDate is null (item not yet purchased), a latest interval can't be
     // calculated, so in that case set latestInterval to the current purchaseInterval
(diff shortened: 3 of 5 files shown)


src/components/ShoppingList/ShoppingList.js
```diff
@@ -62,6 +62,90 @@ function ShoppingList({ listId }) {
...
+        currentDate
+          .diff(DateTime.fromSeconds(item.createdAt.seconds), ['days'])
+          .as('days') >=
+        2 * item.purchaseInterval
+      );
+    } else return null;
+  };
+
+  const sortListItems = (itemA, itemB) => {
+    // if items being compared are both inactive, or neither is inactive, sort by days until next purchase
```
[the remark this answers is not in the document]
connietran-dev: I see it in your PR. It would be helpful if the explanation were included as a comment as well! The PR explanations are quite helpful. In general, it might be good to make sure they're included as comments as well - even if in short form.

src/components/ShoppingList/ShoppingList.js
```diff
@@ -62,6 +62,90 @@ function ShoppingList({ listId }) {
...
+    if (item.daysToPurchase > 30) return 'not-soon';
+  };
+
+  // format the data from firestore into a sorted, filtered array of items
+  // add new properties: item.status and item.daysToPurchase
+  const itemsToDisplay = listItems?.docs
+    .filter((doc) => new RegExp(filter, 'i').test(doc.data().itemName))
+    .map((doc) => {
+      const item = doc.data();
+      item.id = doc.id;
```
connietran-dev, line 142: Thanks for making a single item since Firestore keeps these separate! This definitely makes it easier for all of us in the long-term

src/components/AddItemForm/AddItemForm.js
```diff
@@ -82,6 +84,7 @@ const AddItemForm = ({ listId }) => {
       purchaseInterval: Number(formValues.purchaseInterval),
       lastPurchaseDate: null,
       numberOfPurchases: 0,
+      createdAt: firebase.firestore.FieldValue.serverTimestamp(),
```
connietran-dev, line 87: Do we want to set the createdAt date in the same way we are setting `lastPurchaseDate` in ShoppingList.js on line 49? I know there was some debate on keeping the times consistent.

```js
    db.collection(`lists/${listId}/items`)
      .doc(item.id)
      .update({
        lastPurchaseDate: new firebase.firestore.Timestamp(
          newPurchaseDate.toSeconds(),
          0,
        ),
        numberOfPurchases: firebase.firestore.FieldValue.increment(1),
        purchaseInterval: newPurchaseInterval,
      })
      .catch((err) => {
        console.log(err);
      });
  };
```

src/components/ShoppingListItem/ShoppingListItem.js
```diff
@@ -17,19 +19,27 @@ const ShoppingListItem = ({ listId, itemId, item, checkAsPurchased }) => {
...
+        className={`label label_check-radio item__label item__label_${item.status}`}
+        htmlFor={`item-${item.id}`}
       >
         {item.itemName}
+        <span className="visually-hidden">
+          {item.status === 'soon' && ' Need to buy soon'}
+          {item.status === 'kind-of-soon' && ' Need to buy kind of soon'}
+          {item.status === 'not-soon' && " Don't need to buy soon"}
+          {item.status === 'inactive' && ' Inactive'}
+        </span>
```
connietran-dev, line 42: It might be helpful to put a comment that summarizes your note in the PR for what this span is doing.

It's really nice that the screenreader reads "Need to buy soon", "Need to buy kind of soon", etc. However, what is the `soon`, `kind-of-soon` at the beginning of each `&&` statement used for?

src/App.css
```diff
@@ -87,3 +87,13 @@ BEM conventions we are using:
...
 }
+
+.visually-hidden {
+  clip: rect(0 0 0 0);
+  clip-path: inset(50%);
+  height: 1px;
+  overflow: hidden;
+  position: absolute;
+  white-space: nowrap;
+  width: 1px;
```
connietran-dev, line 98: I've never seen hiding an element defined with CSS in this way. Is there a benefit for doing this over something like `display: none`?

src/components/ShoppingList/ShoppingList.js
```diff
@@ -62,6 +62,90 @@ function ShoppingList({ listId }) {
     setFilter(e.target.value);
   };
 
+  const getDaysToPurchase = (item) => {
+    let nextPurchaseDate;
+    if (item.lastPurchaseDate?.seconds) {
+      // if the item has been purchased before, next purchase date is [purchaseInterval] days from the lastPurchaseDate
```
connietran-dev, line 68: This is super minor, but for my understanding, why do you put purchaseInterval in brackets throughout your comments? Is it denoting something specific about purchaseInterval?

src/components/ShoppingList/ShoppingList.js
```diff
@@ -62,6 +62,90 @@ function ShoppingList({ listId }) {
...
+  const isItemInactive = (item) => {
+    const currentDate = DateTime.fromSeconds(Math.floor(Date.now() / 1000));
+
+    if (item.lastPurchaseDate?.seconds) {
+      // if an item has been purchased before, compare the time elapsed since lastPurchaseDate to the purchaseInterval
+      return (
+        currentDate
+          .diff(DateTime.fromSeconds(item.lastPurchaseDate.seconds), ['days'])
+          .as('days') >=
+        2 * item.purchaseInterval
```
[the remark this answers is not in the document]
connietran-dev: Ah, I see it in your PR. Again, It would be beneficial to add what you have in the PR as a comment here, especially for the long-term so we remember why we multiplied by 2!

connietran-dev (approved):
Great job putting in all of this complex logic and new functionality. It definitely seems like it was a lot of work!

Overall, most of my comments were questions to better understand the code. 

The notes in the description of the PR were helpful for my understanding, so I would suggest translating them into comments in the code so that we have them in the code for the long-term if we ever decide to revisit our code 6 months from now! Great job all

src/App.css
anderswift: Yep! It's used for situations where you want a screen reader to "see" something, but you don't want to display it visually (perhaps there is a visual design element already in place, but you want to provide an auditory fallback for people who can't see it). Screen readers can't see things styled as display: none, so it's necessarily to use this kind of weird hack. [Here's some more information from the a11y project.](https://www.a11yproject.com/posts/2013-01-11-how-to-hide-content/.)

src/components/AddItemForm/AddItemForm.js
anderswift: I think since we only use the createdAt timestamp in one place (setting it in the database) and don't have a JavaScript Date or Luxon DateTime object already in use with this value, we're fine using the serverTimestamp.

src/components/ShoppingList/ShoppingList.js
anderswift: @connietran-dev I think we were just trying to be clear that we were using it to indicate that it had a kind of dynamic value in the sentence, but I can see how it could actually be confusing mixed with code where [ ]'s mean something specific. Do you have a suggestion for what would read clearer to you? Because I'd be happy to change it to improve comment readability.

src/components/ShoppingList/ShoppingList.js
```diff
@@ -62,6 +62,90 @@ function ShoppingList({ listId }) {
...
+      nextPurchaseDate = DateTime.fromSeconds(
+        item.lastPurchaseDate.seconds,
+      ).plus({ days: item.purchaseInterval });
+    } else if (item.createdAt?.seconds) {
+      // if there's no purchase history, estimate it will be bought [purchaseInterval] days from when item was created
+      // (user provides this info at item creation)
+      nextPurchaseDate = DateTime.fromSeconds(item.createdAt.seconds).plus({
+        days: item.purchaseInterval,
+      });
+    } else return null;
```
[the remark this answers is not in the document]
anderswift: Good catch, because I think nothing? I'm pretty sure that was put in with the intention that we could later check for null when we called the function and handle a situation where there was no lastPurchaseDate or createdAt. But then I don't think we ever added anything in to handle that, so this could be removed.

src/components/ShoppingList/ShoppingList.js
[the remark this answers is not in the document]
anderswift: Ah, yes it's in the AC that this is how inactive items are defined. Should definitely add a comment in here, thanks!

src/components/ShoppingList/ShoppingList.js
```diff
@@ -62,6 +62,90 @@ function ShoppingList({ listId }) {
...
+    if (isItemInactive(item)) return 'inactive';
+    if (item.daysToPurchase < 7) return 'soon';
+    if (item.daysToPurchase >= 7 && item.daysToPurchase <= 30)
+      return 'kind-of-soon';
+    if (item.daysToPurchase > 30) return 'not-soon';
+  };
+
+  // format the data from firestore into a sorted, filtered array of items
+  // add new properties: item.status and item.daysToPurchase
+  const itemsToDisplay = listItems?.docs
```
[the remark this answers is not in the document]
anderswift: 1. I was torn on whether it should be at the top of the component or the bottom so I wouldn't mind moving it, but Nick felt it made sense for it to go right before the JSX. @nick-zanetti what do you think after having another perspective?

2. This is also in the AC, and I think it might help to think of it like this: Let's say you have apples on your list with a purchaseInterval of 8 days and bananas with a purchaseInterval of 15 days. Which should go higher up on the list? You might think apples first, then bananas, since the lower purchaseInterval for apples means you purchase them more often. BUT, what if you just bought apples yesterday so they aren't due for at least a week, and meanwhile you haven't bought bananas for almost 3 weeks so based on your history you're well overdue to buy more? From that perspective, it makes sense to use a combination of the lastPurchaseDate and the purchaseInterval to figure out the next date you're expected to buy a given item. "daysToPurchase" is the number of days from now until that next purchase date, and it is always being recalculated since it changes as time passes.

src/components/ShoppingList/ShoppingList.js
anderswift: I think so, and it seemed the easiest way to pass our results of getdaysToPurchase and getItemStatus to the ShoppingListItem component without needing to calculate them more than once (since we need them in this component for sorting).

src/components/ShoppingListItem/ShoppingListItem.js
anderswift: We thought it made sense to set a property item.status back in the ShoppingList component, since we already needed to figure out which items were inactive there in order to sort. We decided to use the strings 'soon', 'kind-of-soon', 'not-soon' and 'inactive' as the 4 possible values for item.status, and these are returned by the getItemStatus function (lines 128-134 of ShoppingList). The thinking was we could reuse these strings as CSS class modifiers after we passed the item object (with its new status property) to ShoppingListItem. Since we also need to print fuller labels for the benefit of human ears, here we are just checking to see which short little string value is stored in item.status so we know which corresponding full label to print.

src/components/ShoppingList/ShoppingList.js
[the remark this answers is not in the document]
nick-zanetti: 1. I think there are arguments to be made for putting it in either the top of the file or right before the render. Since we are rendering based on `itemsToDispaly`, I thought it made to put it just before that JSX, but I see what you're saying about code readability. Let's move it up top and then put the functions in the order they appear in it, I think that gives the file a nice structure for the future. 

2. So I actually made the mistake of thinking we were sorting by `purchaseInterval` when I first brainstormed this issue. Sorting by the `daysToPurchase` is actually the core functionality of the app as I see it, because it combines the user estimated time interval with actual time that has passed since the purchase, allowing for a "smart" estimate of what should be purchased next 🤯 .

src/components/ShoppingListItem/ShoppingListItem.js
nick-zanetti: We can definitely add a short note here and the other places you mentioned @connietran-dev

src/components/ShoppingList/ShoppingList.js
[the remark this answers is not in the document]
SheilaRMoore: Looking at the code, I think it's a good idea to move it to the top. I also think sorting by daysToPurchase is a good idea.

src/App.css
SheilaRMoore: I had recently ready about this. So I thought it was kinda cool to see it in action.

SheilaRMoore (approved):

Cjstoney (commented):
This is looking pretty good so far. When I create a new list of three items, with each item using a different purchase-interval, they are all assigned the class of 'item__label_kind_of_soon' and given the background color of yellow. My understanding of the AC was that each of these should have a distinct visual identifier for the purchaseInterval. I am not sure if that was applied to products that have not been purchase yet though? 
image
image

anderswift:
> This is looking pretty good so far. When I create a new list of three items, with each item using a different purchase-interval, they are all assigned the class of 'item__label_kind_of_soon' and given the background color of yellow. My understanding of the AC was that each of these should have a distinct visual identifier for the purchaseInterval. I am not sure if that was applied to products that have not been purchase yet though?

@Cjstoney The purchaseInterval we have stored in the database represents the app's estimate for how often the user buys an item (ie. every purchaseInterval days you're expected to buy an item). If I'm understanding the AC correctly, the sorting and labelling/colour coding of items is not based on the purchaseInterval directly, but on the "number of estimated days until next purchase" which would change as the next purchase date gets closer. [The conversation here goes into more detail about this distinction and our line of thinking while tackling this issue.](https://github.com/the-collab-lab/tcl-26-smart-shopping-list/pull/43#discussion_r693280204) Of course, I could be misunderstanding the intention in the AC, and/or perhaps our names for things could be worded better to make the difference clearer?

Cjstoney (approved):

src/components/ShoppingList/ShoppingList.js
connietran-dev: I think when I first saw purchaseInterval with brackets, I thought it was an array or maybe an object's property name. I would simply remove it or put it around backtickets to indicate it's code / a variable name. Thanks for being open to suggestion!

src/App.css
connietran-dev: Thank you for linking the article @anderswift . Very helpful and informative!

yenly (approved):
Impressive work! 👏🏽 🚀  I want to echo @connietran-dev's recommendation to include code comments before `getDaysToPurchase()`, `isItemInactive()`, and `sortListItems()` functions  to explain the whys like you had in your PR. Great explainer in PR but future contributors will not always check PR first.

[This is a good pattern to follow](https://github.com/the-collab-lab/tcl-26-smart-shopping-list/blob/main/src/lib/estimates.js#L1-L10).

src/App.css
yenly: TIL 👀

src/components/ShoppingList/ShoppingList.js
```diff
@@ -62,6 +62,90 @@ function ShoppingList({ listId }) {
...
+    } else if (item.createdAt?.seconds) {
+      // if an item has never been purchased, do the same calculating using createdAt date instead of lastPurchaseDate
+      return (
+        currentDate
+          .diff(DateTime.fromSeconds(item.createdAt.seconds), ['days'])
+          .as('days') >=
+        2 * item.purchaseInterval
+      );
+    } else return null;
+  };
```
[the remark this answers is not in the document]
yenly: You can also use the [logical nullish assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_nullish_assignment), `??=` but I haven't seen it in codebases yet.

src/components/ShoppingList/ShoppingList.js
```diff
@@ -62,6 +62,90 @@ function ShoppingList({ listId }) {
...
+      return itemA.itemName.localeCompare(itemB.itemName, 'en', {
+        sensitivity: 'base',
+        ignorePunctuation: true,
+      });
+    } else {
+      // if one item is inactive and the other is not, bump down the inactive item
+      if (itemA.status === 'inactive') return 1;
+      if (itemB.status === 'inactive') return -1;
+    }
+  };
```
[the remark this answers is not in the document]
anderswift: Ooh, `switch (true) { ... }` is such a cool pattern I've never seen before! 🤯 I'm definitely going to remember that one.

PR #44, opened 2021-08-18
Delete shopping item

### LINKED ISSUE #11: 11. As a user, I want to be able to delete items from my shopping list so that my list isn’t cluttered with items I don’t want to buy in the future.
It’s possible our users may make an error when entering an item or may decide they won’t be buying a certain item again in the future. Let’s help them keep their lists tidy!

AC:

* User is able to delete an item from the shopping list
* Before deleting, prompt the user to confirm that they really want to delete the item to prevent accidental deletions
* Deletion should cause the associated record(s) in the database to be deleted

Notes:

* As noted in the wireframe https://github.com/the-collab-lab/smart-shopping-list/blob/main/_resources/wireframes/list-populated.png activating the "Delete" link should pop-up a confirmation dialog because this is a "destructive" action.
* Many teams jump to a modal as the solution here, but modals are fraught with accessibility concerns, so consider using a simple JavaScript confirm dialog.

connietran-dev:
## Description

We've added the ability to delete a Shopping Item by adding a "Delete" button next to each item. A modal appears if you attempt to delete an item asking you to confirm that you'd like to delete it. If you select to delete the item, it removes the item from Firestore, and there is a handleCloseModal to close the modal.

We've also tried to make this modal accessible by allowing you to navigate to it and use it with your keyboard. We've used the following as resources for developing accessible modals:
- Aria practices: https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/dialog.html
- Keyboard navigation: https://tinloof.com/blog/how-to-create-an-accessible-react-modal/

## Related Issue

Closes #11 : As a user, I want to be able to delete items from my shopping list so that my list isn’t cluttered with items I don’t want to buy in the future.

## Acceptance Criteria

- User is able to delete an item from the shopping list
- Before deleting, prompt the user to confirm that they really want to delete the item to prevent accidental deletions
- Deletion should cause the associated record(s) in the database to be deleted

## Type of Changes

### Before

image

### After

image

Functionality:
1. Click "Delete" next to an item.
2. In the modal, clicking "No, Cancel" will close the modal.
3. In the modal, clicking "Yes, Delete" will remove the item from the Shopping List.
4. The item is no longer in Firestore.

Accessibility steps:
1. You should be able to use your keyboard to navigate and select "Delete".
2. When the modal opens, the focus will be on "No, Cancel".
3. You can tab in the modal which will navigate you to either "No, Cancel" or "Yes, Delete". 
4. You can use Enter to select your option.
5. You can escape to close the modal.

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -87,3 +87,11 @@ BEM conventions we are using:
   margin: 0;
   list-style: none;
 }
+
+.display-block {
+  display: block;
+}
+
+.display-none {
+  display: none;
+}
--- a/src/App.css
+++ b/src/App.css
@@ -87,11 +87,3 @@ BEM conventions we are using:
   margin: 0;
   list-style: none;
 }
-
-.display-block {
-  display: block;
-}
-
-.display-none {
-  display: none;
-}
--- a/src/App.js
+++ b/src/App.js
@@ -87,6 +87,7 @@ function App() {
   }
 
   const handleModalOpen = (item, itemId) => {
+    // set item to be deleted to item object and set itemId since it's separate from Firestore
     setItemToDelete({ ...item, id: itemId });
     setShowModal(true);
   };
@@ -100,7 +101,6 @@ function App() {
       .doc(itemToDelete.id)
       .delete()
       .then(() => {
-        console.log(`Document ${itemToDelete.id} successfully deleted`);
         handleModalClose();
       })
       .catch((err) => {
--- a/src/App.js
+++ b/src/App.js
@@ -13,17 +13,20 @@ import { db } from './lib/firebase.js';
 // External Files
 import './App.css';
 
-// View Components
+// Components
 import ListView from './pages/ListView/ListView';
 import AddItemView from './pages/AddItemView/AddItemView';
 import NotFound from './pages/NotFound/NotFound';
 import Home from './pages/Home/Home';
+import Modal from './components/Modal/Modal.js';
 
 // Functions
 import getToken from './lib/tokens';
 
 function App() {
   const [listId, setListId] = useState(null);
+  const [showModal, setShowModal] = useState(false);
+  const [itemToDelete, setItemToDelete] = useState({});
 
   function createList() {
     const token = getToken();
@@ -83,6 +86,28 @@ function App() {
     });
   }
 
+  const handleModalOpen = (item, itemId) => {
+    setItemToDelete({ ...item, id: itemId });
+    setShowModal(true);
+  };
+
+  const handleModalClose = () => {
+    setShowModal(false);
+  };
+
+  const deleteItem = () => {
+    db.collection(`lists/${listId}/items`)
+      .doc(itemToDelete.id)
+      .delete()
+      .then(() => {
+        console.log(`Document ${itemToDelete.id} successfully deleted`);
+        handleModalClose();
+      })
+      .catch((err) => {
+        console.error('Error removing document: ', err);
+      });
+  };
+
   // on component mounting, look for token in local storage and use it to retrieve the list id
   useEffect(() => {
     const token = localStorage.getItem('token');
@@ -100,7 +125,10 @@ function App() {
 
   return (
     <Router>
-      <div className="App container">
+      <div
+        className="App container"
+        aria-hidden={showModal} // if showModal is true, hide the rest of the app
+      >
         <Switch>
           <Route exact path="/">
             {listId ? (
@@ -110,7 +138,11 @@ function App() {
             )}
           </Route>
           <Route path="/list">
-            {!listId ? <Redirect exact to="/" /> : <ListView listId={listId} />}
+            {!listId ? (
+              <Redirect exact to="/" />
+            ) : (
+              <ListView listId={listId} handleModalOpen={handleModalOpen} />
+            )}
           </Route>
           <Route path="/add">
             {!listId ? (
@@ -122,6 +154,13 @@ function App() {
           <Route component={NotFound} />
         </Switch>
       </div>
+
+      <Modal
+        showModal={showModal}
+        handleModalClose={handleModalClose}
+        deleteItem={deleteItem}
+        item={itemToDelete}
+      />
     </Router>
   );
 }
--- a/src/components/Modal/Modal.js
+++ b/src/components/Modal/Modal.js
@@ -1,19 +1,57 @@
+import { useEffect, useRef } from 'react';
 import './Modal.css';
 
 const Modal = ({ showModal, handleModalClose, deleteItem, item }) => {
   const toggleModalClassName = showModal ? 'display-block' : 'display-none';
 
+  const cancelRef = useRef();
+  const deleteRef = useRef();
+
+  useEffect(() => {
+    if (showModal) {
+      // when modal opens, add eventListeners and put initial focus on "No, Cancel"
+      document.addEventListener('keydown', handleKeyEvents);
+      cancelRef.current.focus();
+    } else {
+      document.removeEventListener('keydown', handleKeyEvents);
+    }
+  }, [showModal]);
+
+  const handleKeyEvents = (e) => {
+    // close modal if user hits Escape (27)
+    if (e.keyCode === 27) {
+      handleModalClose();
+    }
+
+    // keep keyboard focus in modal
+    if (e.keyCode === 9) {
+      // if user hits Tab (9)...
+      // and NOT shift while on Delete, put focus on Cancel
+      if (!e.shiftKey && document.activeElement === deleteRef.current) {
+        cancelRef.current.focus();
+        return e.preventDefault();
+      }
+
+      // and shift while on Cancel, put focus on Delete
+      if (e.shiftKey && document.activeElement === cancelRef.current) {
+        deleteRef.current.focus();
+        e.preventDefault();
+      }
+    }
+  };
+
   return (
     <div className={`dialog-backdrop ${toggleModalClassName}`}>
       <div role="alertdialog" aria-modal="true" aria-labelledby="dialog_label">
         <h3 id="dialog_label">{`Are you sure you want to delete ${item.itemName}?`}</h3>
-        <button type="button" onClick={handleModalClose}>
+        <button type="button" onClick={handleModalClose} ref={cancelRef}>
           No, Cancel
         </button>
         <button
           type="button"
           onClick={deleteItem}
           aria-controls={`item-${item.id}`}
+          ref={deleteRef}
         >
           Yes, Delete
         </button>
--- a/src/components/Modal/Modal.js
+++ b/src/components/Modal/Modal.js
@@ -8,7 +8,8 @@ const Modal = ({ showModal, handleModalClose, listId, itemId }) => {
       .doc(itemId)
       .delete()
       .then(() => {
-        console.log('Document successfully deleted');
+        console.log(`Document ${itemId} successfully deleted`);
+        handleModalClose();
       })
       .catch((err) => {
         console.error('Error removing document: ', err);
@@ -17,9 +18,25 @@ const Modal = ({ showModal, handleModalClose, listId, itemId }) => {
 
   return (
     <div className={toggleModalClassName}>
-      <h2>Are you sure you want to delete this item?</h2>
-      <button onClick={handleModalClose}>No, Cancel</button>
-      <button onClick={deleteItem}>Yes, Delete</button>
+      <div
+        role="alertdialog"
+        aria-modal="true"
+        aria-labelledby="dialog_label"
+        aria-describedby="dialog_desc"
+      >
+        <h2 id="dialog_label">Confirmation</h2>
+        <h3 id="dialog_desc">Are you sure you want to delete this item?</h3>
+        <button type="button" onClick={handleModalClose}>
+          No, Cancel
+        </button>
+        <button
+          type="button"
+          onClick={deleteItem}
+          aria-controls={`item-${itemId}`}
+        >
+          Yes, Delete
+        </button>
+      </div>
     </div>
   );
 };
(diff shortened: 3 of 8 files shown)


src/components/Modal/Modal.js
```diff
@@ -0,0 +1,64 @@
...
+        return e.preventDefault();
+      }
+
+      // ...and shift while on Cancel, put focus on Delete
+      if (e.shiftKey && document.activeElement === cancelRef.current) {
+        deleteRef.current.focus();
+        return e.preventDefault();
+      }
+    }
+  };
```
nick-zanetti, line 40: Really impressive stuff here, everything in this useEffect is new to me conceptually but you really commented it well and made it easier to understand.

src/components/ShoppingListItem/ShoppingListItem.js
```diff
@@ -2,7 +2,12 @@ import { useState, useEffect } from 'react';
 
 import isUnder24hSincePurchased from '../../utils/isUnder24hSincePurchased.js';
 
-const ShoppingListItem = ({ listId, itemId, item, checkAsPurchased }) => {
+const ShoppingListItem = ({
+  itemId,
+  item,
+  checkAsPurchased,
+  handleModalOpen,
```
nick-zanetti, line 9: Passing this down so many levels makes me uncomfortable, but since the app if functionally finished at this point, I think it's fine.

nick-zanetti (approved):
Great job! Even though modals have a lot of accessibility concerns, you seem to have handled all of them really well. All the AC look fulfilled to me. 

I am a little curious as to why you ultimately went with the modal vs. a confirm dialog, since the latter seems like it might have been less work? I assume we'll discuss on Sunday.

src/components/Modal/Modal.js
connietran-dev: Yes! If you comment this code out and use a keyboard to open the modal, if you start tabbing or shift/tabbing into the modal, the keyboard focus will start leaving the modal into the shopping list behind it.

src/components/ShoppingListItem/ShoppingListItem.js
connietran-dev: This was definitely a struggle! We originally had most of this modal code right at ShoppingList. But because we add an `aria-hidden=true` to hide the App when the modal is open, we had to move everything up to App.js. We can definitely  consider refactoring to use React Context in the future so we don't have to prop drill something like this so many levels down.

src/App.js
```diff
@@ -83,6 +86,28 @@ function App() {
...
+      .doc(itemToDelete.id)
+      .delete()
+      .then(() => {
+        handleModalClose();
+      })
+      .catch((err) => {
+        console.error('Error removing document: ', err);
+      });
+  };
+
```
anderswift, line 110: See my other comment on Modal.js about event listeners getting added too many times. I suggested there that one potential fix would be to split the handleKeyEvents function up, connecting the tab/tab+shift handling to onKeyDown events on the buttons, and then moving the escape key stuff into a separate function here in App.js. Why move it to App.js? So we can wrap it in a useCallback with no dependencies and ensure we only have one version of the function. (In Modal.js handleModalClose would be a dependency because it's passed down as a prop.)

I put sample code for the first part in my other comment, and here's the escape key portion I tested if you want to try it out. (Note that you would need to import useCallback at the top of this file, so `import { useState, useEffect, useCallback } from 'react';`.)

```
  // close modal if user hits Escape (27)
  const closeModalOnEsc = useCallback((e) => {
    if (e.keyCode === 27) handleModalClose();
  }, []);

  useEffect(() => {
    if(showModal) {
      document.addEventListener('keydown', closeModalOnEsc);
    } else {
      document.removeEventListener('keydown', closeModalOnEsc);
    }
  }, [showModal, closeModalOnEsc]);
```
Optional addition: we could even put in the "close when backdrop clicked" feature we didn't have time to finish up pretty easily by expanding on the above code like so...

```
  // close modal if user hits Escape (27)
  const closeModalOnEsc = useCallback((e) => {
    if (e.keyCode === 27) handleModalClose();
  }, []);

  // close modal if user hits Escape (27)
  const closeModalOnClickAway = useCallback((e) => {
    if (e.target.classList.contains('dialog-backdrop')) handleModalClose();
  }, []);

  // when modal opens/closes, add/remove event listeners
  useEffect(() => {
    if(showModal) {
      document.addEventListener('keydown', closeModalOnEsc);
      document.addEventListener('click', closeModalOnClickAway);
    } else {
      document.removeEventListener('keydown', closeModalOnEsc);
      document.removeEventListener('click', closeModalOnClickAway);
    }
  }, [showModal, closeModalOnEsc, closeModalOnClickAway]);
```

src/components/ShoppingListItem/ShoppingListItem.js
```diff
@@ -15,22 +20,32 @@ const ShoppingListItem = ({ listId, itemId, item, checkAsPurchased }) => {
...
+        htmlFor={`item-input-${itemId}`}
       >
         {item.itemName}
       </label>
+
+      <button
+        type="button"
+        onClick={() => handleModalOpen(item, itemId)}
+        aria-controls={`item-${itemId}`} // destructive delete controls shopping list item id
+        aria-label={`Delete ${item.itemName}`}
```
anderswift, line 44: Screen reader reads this out perfectly! Very good UX in JAWS and NVDA. 👍

src/components/ShoppingListItem/ShoppingListItem.js
anderswift: @nick-zanetti Maybe we can make an issue for refactoring some things into React context?

anderswift (commented):
I love that you decided to take on the tricky task of building an accessible modal from scratch! The screen readers I tested did really well with the way you structured the UI, reading out helpful text at all times—good job figuring out the aria attributes! 

It would be great for both screen reader and keyboard users if we could find a way to add that one missing detail: returning focus to the right place in the list after the modal closes. I also noticed that extra event handlers are accumulating as the modal opens/closes, so made a couple suggestions for a possible fix. Happy to help work through any of that, and maybe we can get some mentor feedback as well.

src/App.js
```diff
@@ -100,7 +125,10 @@ function App() {
 
   return (
     <Router>
-      <div className="App container">
+      <div
+        className="App container"
+        aria-hidden={showModal} // if showModal is true, hide the rest of the app
```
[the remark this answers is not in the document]
abowler2: While this does work to hide the element and as such keeps the screen reader from hitting it, I would caution against this approach as the component still renders and any logic within it will run.  

As an example, I added a `console.log` within the Modal component and as you can see below that fired and logged to the console with the hidden attribute vs conditionally rendering the component using `showModal && <Modal />`

| aria-hidden | conditional render |
| ------------- | ------------- |
| <video src="https://user-images.githubusercontent.com/19564687/130313275-c0b568cd-f23d-4640-930f-3d225766c428.mov" />  | <video src="https://user-images.githubusercontent.com/19564687/130313301-ded67317-8e5a-4a24-81c4-19f219d3d017.mov" />  |

src/components/Modal/Modal.js
yenly, line 40: By moving the `handleKeyEvents` function into the `useEffect` and adding a clean up function, we ensure only we are only adding the eventlistener when it's needed and removing it when the modal component unmounts. Here's is [the recommendation on how to handle 'exhaustive-deps' lint rule](https://github.com/facebook/react/issues/14920).
```suggestion
  useEffect(() => {
    const handleKeyEvents = (e) => {
      // close modal if user hits Escape (27)
      if (e.keyCode === 27) {
        handleModalClose();
      }

      // this keeps keyboard focus within modal. if user hits Tab (9)
      if (e.keyCode === 9) {
        // ...and NOT shift while on Delete, put focus on Cancel
        if (!e.shiftKey && document.activeElement === deleteRef.current) {
          cancelRef.current.focus();
          return e.preventDefault();
        }

        // ...and shift while on Cancel, put focus on Delete
        if (e.shiftKey && document.activeElement === cancelRef.current) {
          deleteRef.current.focus();
          return e.preventDefault();
        }
      }
    };
    if (showModal) {
      // when modal opens, add eventListeners and put initial focus on "No, Cancel"
      document.addEventListener('keydown', handleKeyEvents);
      cancelRef.current.focus();
    }
    return () => document.removeEventListener('keydown', handleKeyEvents);
  }, [handleModalClose, showModal]);
```

src/App.css
```diff
@@ -87,3 +87,11 @@ BEM conventions we are using:
...
   list-style: none;
 }
+
+.display-block {
+  display: block;
+}
+
+.display-none {
+  display: none;
+}
```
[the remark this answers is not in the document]
yenly: +1 on meaningful naming.

src/App.js
```diff
@@ -83,6 +86,28 @@ function App() {
     });
   }
 
+  const handleModalOpen = (item, itemId) => {
+    // set item to be deleted to item object and set itemId since it's separate from Firestore
+    setItemToDelete({ ...item, id: itemId });
+    setShowModal(true);
```
yenly, line 92: This is an interesting approach to save the item object for deletion in App state before doing so.

src/components/ShoppingListItem/ShoppingListItem.js
nick-zanetti: @anderswift @connietran-dev I actually think that would be more work than it's worth, since the app is now essentially done functionally.

src/App.js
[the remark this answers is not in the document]
yenly: This is an interesting approach. I'm not sure what the right way to approach this as well. Will have to research it.

yenly (approved):
👏🏽 I'm impressed with your choice in creating a custom dialog component and making it keyboard navigation-friendly. 🚀

abowler2 (approved):
To second what @yenly said, really nice work taking on the modal for this issue.

src/App.js
Cjstoney: This is more of a question for clarity on my end, but why is the ID "separate from Firestore"? It seems we need to associate the id with an item in the Firestore in order to delete it, but it is not coming from the item object itself?

Cjstoney (approved):
This works really well and is quite impressive, especially considering it was built from scratch and works with keyboard navigation. Good work!

anderswift (approved):

src/components/Modal/Modal.js
anderswift: ☝ This is a much simpler solution than the one I worked out.

src/App.js
anderswift: @Cjstoney The structure of the data returned from Firestore has the document id "separate" from the field data (`doc.id` has the id and `doc.data()` returns an object with the other fields and values). We originally didn't do anything to combine these values and just used item and itemId to hold these different bits from Firestore. 

This sprint Nick and I decided to reorganize the data returned from Firestore before passing it down through the app, so that we'd have a simpler, cleaner array of items with item.id and some other helpful properties we needed. We coordinated with Connie & Sheila about this coming change, and that may have informed this line if I'm remembering correctly, since `setItemToDelete({...item, id: itemId})` would work with their current code but be really easy to simplify to `setItemToDelete(item)` once we merge our branches together and no longer have a separate itemId prop.

src/App.js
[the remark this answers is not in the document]
anderswift: @abowler2 That is true! Hmm. For modals, drop down menus, etc. I've always done them this same way, having them render (or just be included in markup if vanilla JS) but shown/hidden with a CSS class, so later I could use CSS transitions to animate them fading or sliding in and out of view. The bootcamp I attended taught this method as well. Is this basically bad for performance in React? And is there a good alternate way you can still have an animation with the `showModal && <Modal />` approach?

src/components/ShoppingListItem/ShoppingListItem.js
anderswift: @nick-zanetti I don't think it's necessary, but might be a good opportunity to practice with React context if anyone has some time and wants to give it a try. Refactoring can be a really great learning experience.

src/App.js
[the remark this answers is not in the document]
anderswift: @yenly I think it comes from [this w3.org page](https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/dialog.html). From what we could determine, the aria-modal attribute on the modal alone *should* take care of hiding things that are not the modal, but it is new so not fully supported everywhere. For full support, it's suggested to use aria-hidden="true" on "each element containing a portion of the inert layer" and that the modal must not be inside an aria-hidden layer. I found this general pattern pretty common: 
```
<div aria-hidden="true">the whole rest of the page</div>
<div aria-modal="true">the active modal</div>
```
Though what we weren't certain of is what percentage of screen reader users might have legacy software that doesn't fully support aria-modal, therefore how necessary aria-hidden really is!

src/components/ShoppingListItem/ShoppingListItem.js
connietran-dev: I've logged #45 as a future enhancement/issue we could take on 😃

src/App.css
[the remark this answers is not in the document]
connietran-dev: This is a great suggestion. We've made [a new commit](https://github.com/the-collab-lab/tcl-26-smart-shopping-list/pull/44/commits/75679e4912f07995693228cf319afa36d79a9e06) to rename these CSS classes to `.dialog-open` and `.dialog-close` so that they're related to the dialog and we can keep adding CSS to them. We've moved them into Modal.js. For now, we didn't add just a `.dialog` class but we certainly can in the future

src/App.js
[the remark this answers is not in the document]
connietran-dev: Thank you all for the helpful input, especially April with the gifs (very helpful and clear what the issue is.) Is this something we can discuss Sunday if there is time?

src/components/Modal/Modal.js
connietran-dev: Committed your suggestion Yenly. Thank you all for your input and help in solving this, including @anderswift ! I'm definitely learning a lot about React hooks and the Effect hook through TCL. Thank you! <3

PR #66, opened 2021-08-27
Add 'share token' component to header

### LINKED ISSUE #56: Create component for list sharing and add to Header
Add the component for list sharing to the Header.js component (see issue #53)

anderswift:
## Description

This PR adds functionality allowing the user to view and copy their list token for the purpose of sharing their list with another user. A new component ShareToken was created for this purpose and added to the Header component, as in the wireframe. We also did some minor refactoring of App.js to allow for storing the token in state and passing it down to the component.

### Token Copy button
The token is presented in a readonly form field, so makes it easy to copy. We also created a 'Copy' button which copies the token to the user's clipboard.

### Responsive Layout

The ShareToken component contains a toggle button for mobile screens, which will allow it to be shown/hidden by the user as needed when there is limited screen space. On larger screens, the toggle button disappears and the 'Share token' area is always visible.

### Some CSS refactoring

Since the token field with copy button is similar to the filter with clear button, a reusable block called `.form-group` was created, and styles positioning the field and button moved from the `.filter` block to the more generic `.form-group`.  The App.css file was also reordered a bit to try and better organize the rules, grouping similar elements and moving more commonly used ones closer to the top of the file.

## Related Issue

Closes #56 

## Type of Changes

## Updates

image

share

1. Pull down the branch and install dependencies if you haven't already.
2. Run npm start.
3. With a narrow browser window, test toggling the 'share' area open and closed.
4. On a wider browser window, the 'share area' should always be visible as in the screenshot provided.
5. Test to see if the token is correctly displayed.
6. Test that the copy button successfully copies your token to the clipboard (should work in most browsers, we checked on recent versions of Firefox, Chrome and Edge).


src/App.css
```diff
@@ -275,3 +178,119 @@ CSS Organization reference: https://gist.github.com/awkale/ad46e2ade70e833fa178
...
+@media only screen and (min-width: 1200px) {
+  .error_type_field {
+    font-size: 1.25rem;
+  }
+}
+
+/* groups together elements to allow for buttons in text field */
+.form-group {
+  position: relative;
+}
```
connietran-dev, line 235: I appreciate having this comment!

src/App.css
```diff
@@ -275,3 +178,119 @@ CSS Organization reference: https://gist.github.com/awkale/ad46e2ade70e833fa178
...
+  .error_type_field {
+    font-size: 1.125rem;
+  }
+}
+
+@media only screen and (min-width: 1200px) {
+  .error_type_field {
+    font-size: 1.25rem;
+  }
+}
```
connietran-dev, line 230: Do we want to consider making these min-width into CSS variables in case they ever change?

It also might be good to start putting comments in App.css to group the variables together with headings. I appreciate the comments in App.css explaining the purpose of the variables!

src/components/Header/Header.css
```diff
@@ -1,5 +1,19 @@
 .header {
-  text-align: center;
+  display: grid;
+  grid-template:
+    'logo share-toggle'
+    'share-token share-token';
```
connietran-dev, line 5: What is `grid-template` doing here? This is a new concept/CSS property for me!

src/components/ShareToken/ShareToken.js
```diff
@@ -0,0 +1,58 @@
...
+      document.execCommand(token); // possible fallback for older browsers
+      console.log(err);
+    });
+  }
+
+  return (
+    <>
+      <button
+        type="button"
+        onClick={() => setShowMobileShare(!showMobileShare)}
```
connietran-dev, line 21: I'm really liking the ability for it to toggle the hiding and showing on every click!

src/components/ShareToken/ShareToken.js
```diff
@@ -0,0 +1,58 @@
...
+        className="icon-only-button header__share-toggle-button share-toggle-button"
+        aria-label="Share list"
+      >
+        <ShareIcon aria-hidden="true" focusable="false" />
```
connietran-dev, line 25: I'm assuming this is for desktop?

src/components/ItemFilter/ItemFilter.css
```diff
@@ -1,23 +1,7 @@
...
-  position: absolute;
-  bottom: var(--border-width);
-  right: var(--border-width);
-  width: 2.75rem;
-  height: 2.75rem;
-  font-size: inherit;
-  color: var(--charcoal);
-  border: none;
-}
-
```
connietran-dev, line 20: Why were these filter styles removed or where did they get moved to?

connietran-dev (approved):
I'm not 100% sure I understand what the refactor for App.js is doing, but the functionality works great! I love the mobile toggling and copying on desktop works great. I'm glad we're able to add this feature to our app!

src/App.css
anderswift: @connietran-dev Since these are media queries (and pretty standard breakpoints) I hesitate to make them variables because that would encourage us to think of them as easily changeable. Changing the media query breakpoint would require testing every single place it occurs in the browser again! If we have to change them, I'd rather just search replace and do it very mindfully while extensively checking my layout.

Re: putting headings, I totally agree some of the CSS files are getting unwieldy. The only reason I haven't put in header type comments is because I'm planning to ask in office hours about the affect on performance of splitting them into different files with @imports, because I'm not 100% sure how React handles CSS in the end build. But we will definitely come back to that and do something to break it up better.

src/components/ItemFilter/ItemFilter.css
anderswift: @connietran-dev Because I used a "button in form field" style thing again for the copy button, I decided to put those styles in a reusable block instead of specific to the filter, that we I could use the same block for both and if we change it only change it once. The "button in form field" combo block is called `.form-group` with `.form-group__button` and `.form-group__field`. I put a comment above that section explaining the purpose of `.form-group, though open to suggestions on better naming!

src/components/Header/Header.css
anderswift: Oh, I love grid-template! Personally, I sometimes find it a clearer way to define a grid and understand exactly what will go where. [Here's an explainer](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/Grid_Template_Areas), but basically the words are defining areas of the grid almost like a table (so yeah they need that line break even though it's weird-looking in CSS):
|            |  col 1  |  col 2     |
|--------|:-------:|----------:|
| row 1 |   logo   |  share-toggle  |
| row 2 |  share-token  |  share-token  |

By defining that grid-template, you've giving names to areas of the grid, and then you can do the following to tell CSS where to stick something: 
```
.header__share-toggle-button {
  grid-area: share-toggle;
}
```
Pretty cool, huh?

src/components/ShareToken/ShareToken.js
anderswift: Just wait 'til we get fancy and animate everything!

src/components/ShareToken/ShareToken.js
anderswift: Not following you here. Which thing? This looks like the button that is visible on mobile and allows you to toggle the token share area open and closed. I can tell you that the `aria-hidden="true" focusable="false"` is a pattern I see used for inline SVG icons. You hide them from possibly older screen readers that might get confused, since we already have an aria-label that is meaningful and the icon is just decoration. And you prevent focus because apparently certain crappy browsers allow focus on SVGs, which we don't want—we want that focus on the button, or other actually interactive stuff!

anderswift:
> I'm not 100% sure I understand what the refactor for App.js is doing, but the rest of the functionality works great!

@connietran-dev Sorry about that, I will put something in the PR description. I added that form-group class to make the button in field a reusable block, and other than that all I did was reorder things to make more logical sense and group similar elements together, and put more commonly used ones closer to the top. We should definitely revisit better organizing App.css!

src/components/ShareToken/ShareToken.js
```diff
@@ -0,0 +1,58 @@
...
+          />
+          <button
+            type="button"
+            className="share-token__copy-button icon-only-button form-group__field-button"
+            onClick={copyToken}
+          >
+            <CopyIcon aria-hidden="true" focusable="false" />
+          </button>
+        </div>
+      </div>
```
anderswift, line 53: We should add in a few more aria attributes here to show when the button is expanded, as well as a `role="region"` I think on the `div.share-token`. Will make a note to test with a screen reader tomorrow and make those a11y tweaks!

src/components/ShareToken/ShareToken.js
```diff
@@ -0,0 +1,58 @@
+import { useState } from 'react';
+
+import { ReactComponent as ShareIcon } from '../../images/icon-share.svg';
+import { ReactComponent as CopyIcon } from '../../images/icon-copy.svg';
```
nick-zanetti, line 4: This is a nice feature! I feel like I learn something on every PR lately.

src/components/ShareToken/ShareToken.js
```diff
@@ -0,0 +1,58 @@
...
+
+const ShareToken = ({ token }) => {
+  const [showMobileShare, setShowMobileShare] = useState(false);
+
+  function copyToken() {
+    navigator.clipboard.writeText(token).catch((err) => {
+      document.execCommand(token); // possible fallback for older browsers
+      console.log(err);
+    });
+  }
```
nick-zanetti, line 15: What do we think of adding some feedback for the user when they click the copy button? Should be pretty easy to do with the below logic, followed by an additional `<label>` on the `shareToken` input (I've put this second part in a separate comment at the relevant spot below). 

```suggestion
const ShareToken = ({ token }) => {
  const [showMobileShare, setShowMobileShare] = useState(false);
  const [copySuccess, setCopySuccess] = useState(false);

  function copyToken() {
    navigator.clipboard.writeText(token).catch((err) => {
      document.execCommand(token); // possible fallback for older browsers
      console.log(err);
    });
    setCopySuccess(true);
  }
```

src/components/ShareToken/ShareToken.js
```diff
@@ -0,0 +1,58 @@
...
+            readOnly
+          />
+          <button
+            type="button"
+            className="share-token__copy-button icon-only-button form-group__field-button"
+            onClick={copyToken}
+          >
+            <CopyIcon aria-hidden="true" focusable="false" />
+          </button>
+        </div>
```
nick-zanetti, line 52: Here is how we could do a quick label for a feedback message that the text was copied successfully (continued from my comment above). This is pretty bare bones but I assume we could use CSS to make this transition in nicely and maybe even in a speech bubble style like you see when you click on Github's copy buttons. 

```suggestion
        </div>
         {copySuccess ? (
          <label htmlFor="shareToken">Copied!</label>
        ) : (
          <label hidden htmlFor="shareToken">
            Copied!
          </label>
        )}
```

nick-zanetti (approved):
Looks great! I do think we should add some kind of feedback message (that says "copied!" or something) when the user clicks the copy button. That can just be for a future round of styling/ etc. though, and we may change how we decide to tackle it based on broader styling choices.

src/components/ShareToken/ShareToken.js
anderswift: I just learned you could do it this way! Super handy, and let's you micromanage what's going on with the style of the different parts of the SVG if you put classes on the different shapes.

src/components/ShareToken/ShareToken.js
anderswift: Ooh, fantastic idea, I love it!

anderswift: @nick-zanetti Thanks for the reminder earlier today that github uses this kind of feature. I ended up looking at their markup and CSS to see how they did it (was curious about how they handled a11y), so slightly modified approach but inspired by your suggested code. Check it out!

PR #67, opened 2021-08-27
Basic modal style

### LINKED ISSUE #59: Add basic CSS for modal

anderswift:
## Description

This PR uses our global CSS classes/variables and adds basic styles to the modal to match the wireframe. It also add an additional event listener to close the modal when the user clicks outside it.

## Related Issue

Closes #59 

## Type of Changes

### Before

image

### After

image
modal

1. Pull down the branch and install dependencies if you haven't already.
3. Open the modal by clicking the delete button for an item. Check whether it matches the screenshot. 
4. Close and open the modal, and check that it fades in and out smoothly.
5. Try closing the modal by clicking outside of the box on the backdrop overlay.

### DIFF
--- a/src/components/Modal/Modal.js
+++ b/src/components/Modal/Modal.js
@@ -46,10 +46,11 @@ const Modal = ({ showModal, handleModalClose, deleteItem, item }) => {
         aria-labelledby="dialog-label"
       >
         <h3 className="dialog__heading" id="dialog-label">
-          {`Are you sure you want to delete ${item.itemName}?`}
+          Are you sure you want to delete{' '}
+          <strong className="dialog__item-name">{item.itemName}</strong>?
         </h3>
         <button
-          className="dialog__cancel button"
+          className="dialog__button button"
           type="button"
           onClick={handleModalClose}
           ref={cancelRef}
@@ -57,7 +58,7 @@ const Modal = ({ showModal, handleModalClose, deleteItem, item }) => {
           No, Cancel
         </button>
         <button
-          className="dialog__delete button button_type_delete"
+          className="dialog__button button button_type_delete"
           type="button"
           onClick={deleteItem}
           aria-controls={`item-${item.id}`} // destructive delete controls shopping list item id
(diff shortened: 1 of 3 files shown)


src/components/Modal/Modal.js
```diff
@@ -29,30 +29,47 @@ const Modal = ({ showModal, handleModalClose, deleteItem, item }) => {
         }
       }
     };
+
+    const closeOnClickOutside = (e) => {
+      if (e.target.classList.contains('dialog')) {
```
connietran-dev, line 34: I see you changed the className from `dialog-backdrop` to `dialog`. I think it might be helpful if we keep it `dialog-backdrop` to help keep this function clear that when you click on the backdrop/overlay with the 0.3 opacity, it closes the modal. It took me a second to understand what element actually closes the modal. I do see that .dialog has its own CSS as well, so maybe there is still a reason to keep the .dialog class on that overlay div

connietran-dev (approved):

connietran-dev (approved):
The modal is looking good, and I like the feature of clicking outside of the modal to close it. It's also great to learn from your code on how to do it!

src/components/Modal/Modal.js
anderswift: @connietran-dev I'm following BEM conventions to be consistent, so the block is dialog representing the whole set of things, and everything inside is an element of dialog. I don't want to abandon our convention after coming this far, so any suggestions for naming that would still fit? Some ideas:

- `dialog-with-backdrop` and `dialog-with-backdrop__element`
- `dialog-w-backdrop` and `dialog-w-backdrop__element`
- `dialog-backdrop` and `dialog-backdrop__element`
- `dialog-overlay` and `dialog-overlay__element`
- leave it as `dialog` and put a comment in the CSS above it

connietran-dev: Ah, makes sense. We should definitely use .dialog as the whole set. I like dialog-background!

src/components/Modal/Modal.js
```diff
@@ -29,30 +29,47 @@ const Modal = ({ showModal, handleModalClose, deleteItem, item }) => {
         }
       }
     };
+
+    const closeOnClickOutside = (e) => {
```
nick-zanetti, line 33: Good thinking on this, I think this is a really important feature to have. I think this is almost always how I close modals.

nick-zanetti (approved):
This all looks great to me, pretty straightforward!

PR #68, opened 2021-08-27
Uncheck feature

### LINKED ISSUE #40: Modify checkmark logic
Currently, checking the item as purchased will lock that checkbox as disabled for 24 hours. (This is also confusing for screen reader users, as immediately disabling the checkbox causes it to lose focus and then the page is read from the top again.)

A better UX for both situations might be to allow the user to reverse the action during a 5 minute interval after an item is checked. 

This should also address the keyboard skipping over checked items.

If the item is already purchased within 24 hours, then show the message that it's already been purchased.

anderswift:
## Description

This PR adds functionality to handle the user checking an item that is already checked. Previously, checkboxes for items marked as purchased were just disabled. This update allows us to have active checkboxes at all times, which has the benefit of keeping the focus state as a checkbox is checked. Finally, additional notice/error messages are provided to the user when checking/unchecking items. These are set in the new `itemNotice` state, which is an object with properties related to the type and content of the message (including whether it should use aria assertive or polite, whether it should be styled as an error, whether it should be only for screen readers so visually hidden.)

### Unchecking to undo an accidental purchase
When a user checks an item as purchased, there is now a 5 minute window during which they can uncheck the item, allowing them to correct an accidental purchase. To ensure the information doesn't get destroyed when an item is checked off and can therefore be restored, the `checkAsPurchased` function now adds an extra field called `undoRestore` used to save a copy of the previous lastPurchaseDate and purchaseInterval.  

When an item is unchecked, `handleUncheck` runs a function called `isPurchaseWithinUndoWindow` to verify if it's still within the 5 minute "undo window." If it is, the `uncheckAsPurchased` function passed down from `ShoppingList` handles restoring the old data in the Firestore entry for the item. A hidden user noticefor screen readers is then printed saying the purchase was undone.

### Unchecking outside the undo window
If more than 5 minutes has passed, the user is not allowed to uncheck an item, since it's not likely to be accidental and would affect their purchase history tracking. In this case, a user notice is printed saying "You already purchased this in the last 24 hours" so that the user can understand why their action failed. This message appears inline underneath the item in the list to keep things uncluttered and simple to understand, and is destroyed once the checkbox loses focus. This method was chosen over a time out to give the user more control over the length of time the message is shown, with accessibility in mind. 

### New user notices when checking an item as purchased

Since a system for showing item-related messages is now in place, this was a good opportunity to add two extra user feedback messages in the same way. When a user marks an item as purchased, a screen reader only message is printed (ie. "Purchased bananas". If something fails with this operation, a "Sorry, there was a problem" error is printed and available to the user both visually and as a screen reader alert. 

## Related Issue

Closes #40 

## Type of Changes

## Updates

image

image

1. Pull down the branch and install dependencies if you have not already.
3. Try unchecking an item that is already checked. You should see a notice underneath the item.
4. Try checking an item, and then unchecking. You should be able to uncheck if it's within 5 minutes of checking. Be sure to test with multiple items, including new items with no lastPurchaseDate, and items that have been purchased a few times.
5. If you have a screen reader, you can try checking and unchecking items, and listening for the visually hidden user notices.


src/components/ShoppingList/ShoppingList.js
```diff
@@ -168,18 +168,34 @@ function ShoppingList({ listId, handleModalOpen }) {
...
+
+  // restores item's previous lastPurchaseDate, purchaseInterval and numberOfPurchases
+  // for use when a user accidentally checks an item off, and wants to undo
+  const uncheckAsPurchased = (item) => {
+    return db
+      .collection(`lists/${listId}/items`)
+      .doc(item.id)
+      .update({
+        // decrement total purchases by 1
+        numberOfPurchases: firebase.firestore.FieldValue.increment(-1),
```
nick-zanetti, line 194: I was wondering why you didn't use "decrement" then saw on the docs for firestore `FieldValue` that there is no decrement method. I wonder why?

nick-zanetti (approved):
Everything here works for me and I think this was handled well! I don't think I have many comments but we can discuss more on our call today.

connietran-dev (approved):
I appreciate the comments! It is also working for me, not seeing issues

src/components/ShoppingList/ShoppingList.js
```diff
@@ -168,18 +168,34 @@ function ShoppingList({ listId, handleModalOpen }) {
...
-          0,
-        ),
+        lastPurchaseDate: firebase.firestore.FieldValue.serverTimestamp(),
         numberOfPurchases: firebase.firestore.FieldValue.increment(1),
         purchaseInterval: newPurchaseInterval,
-      })
-      .catch((err) => {
-        console.log(err);
+        // back up some info in case user mistakenly checks item and wants to undo
+        undoRestore: {
```
abowler2, line 179: nit: I'm not sure this variable name conveys well what it's purpose is.  Perhaps something like "backupValues" or "originalValues" might read a bit better?

src/components/ShoppingListItem/ShoppingListItem.js
```diff
@@ -1,72 +1,159 @@
...
+  };
 
   // update whether item is recently purchased
   useEffect(() => {
-    // make sure properties exist and are not null
-    if (item?.lastPurchaseDate?.seconds)
-      setRecentlyPurchased(
-        isUnder24hSincePurchased(item.lastPurchaseDate.seconds),
-      );
+    // if last purchase date is null, make sure
```
abowler2, line 82: Did you intend to leave more information here in this comment?

src/components/ShoppingListItem/ShoppingListItem.js
```diff
@@ -1,72 +1,159 @@
...
+          aria-controls={`item-${item.id}`} // destructive delete controls shopping list item id
+          aria-label={`Delete ${item.itemName}`}
+          className="item__delete-button icon-only-button"
+        >
+          <DeleteIcon aria-hidden="true" focusable="false" />
+        </button>
+      </div>
+
+      {/* A space for item-related user feedback, which may be visual or for screen readers only */}
+      <div
```
abowler2, line 147: Since you aren't actually holding space here should this be a conditional render instead of an empty div when no message is present?

abowler2 (approved):
This looks good.  Left just a few small, non-blocking comments

src/components/ShoppingListItem/ShoppingListItem.js
anderswift: @abowler2 I tested that with screen readers and if the aria-live is not present beforehand, the message doesn't seem to get read! Seems like there needs to be an empty div there that the screen reader knows to watch?

src/components/ShoppingList/ShoppingList.js
anderswift: Actually, I agree. I like backupValues and will switch that out before merging.

src/components/ShoppingList/ShoppingList.js
anderswift: Yes, I thought that was kind of quirky as well.

src/components/ShoppingListItem/ShoppingListItem.js
anderswift: @abowler2 Thanks for catching that! I will complete that thought, which I think was meant to say: 
`// if last purchase date is null, make sure to set isRecentlyPurchased to false`
I noticed that the way we were setting (or I believe not setting) isRecentlyPurchased when there was a non-null last purchase date created a bug where unchecking new items didn't work! So had to make a small change here, and wanted to note it.

src/components/ShoppingListItem/ShoppingListItem.js
abowler2: Hmmm.... I didn't experience that but then I'm only testing with VoiceOver so it may be different for other screen readers.  Since there is no logic within that div, it's not a big deal to leave it as is.

anderswift: Just tried it again to double check and yeah, NVDA and Jaws don't like it. 🙄

PR #93, opened 2021-09-03
Add item panel

### LINKED ISSUE #32: Refactor AddItemForm so db is not called on every submit

### LINKED ISSUE #47: Create 'add item' form panel
- Change Add Item link to button.
- Change "Add Item" view to a panel that slides from the right to the left.

Use wireframes: https://drive.google.com/drive/u/2/folders/1N7dg2vNaFqQQxLSSRAvoOXmsspXB_UCJ

anderswift:
## Description

This PR moves the AddItemForm from it's own page to the ListView page, in order to create a sliding panel on the same page as the shopping list. It also refactors the way the listItems data is shared between components, allowing us to reuse the data retrieved from Firestore for both the ShoppingList and the AddItemForm component (which previously fetched a copy of the list every time a user tried to add an item).

This was really tricky, particularly because I wanted to be sure larger font sizes wouldn't get cut off on the add form and the user would still be able to scroll. I ended up using a combination of fixed positioning for a background and absolute positioning in an area I knew would be empty, but allowing any amount of height for the form.

Possibly still some a11y issues at certain small sizes where it functions more like a modal, but not sure how to target JavaScript based on screen size effectively in the way you can CSS.

## Related Issue

Closes #47 
Closes #32 

Merged into this branch are several other related PRs, so also:
Closes #99 
Closes #98 

## Type of Changes

### After

add panel

1. Pull the branch, install dependencies and run npm start.
2. Test opening the add item form at different screen sizes.
3. Try adding some items to make sure it's working properly..


SheilaRMoore (approved):
I took look at this on the computer and on mobile. Looks great on both.  I like how the aria label was added here as well and it worked well with the current screenreader that I have.  Great job on the CSS, especially for mobile. I think the way this was implemented makes the app more user friendly.  Didn't really have any questions that I could think of at the moment.

nick-zanetti (approved):

connietran-dev (approved):

